PR #1, opened 2022-02-04
feat: add button-full-width class to Button component

kamaljohnson:
### Feature
Added class `button-full-width` which when set to true will set the button width to the element width that it is contained in, It `is set to false as default`.

#### How to use
``` html
<Button icon-right="chevron-down" :width-full="true">Button</Button>
```


readme.md
```diff
@@ -102,6 +102,7 @@ app.mount("#app");
 <Button icon-left="menu">Menu</Button>
 <Button icon-right="external-link">Link</Button>
 <Button :loading="true">Loading</Button>
+<Button :button-full-width="true">Full Width Button</Button>
```
NagariaHussain, line 105:
```suggestion
<Button :full-width="true">Full Width Button</Button>
```

Button is redundant anyways, the readability remains same.

NagariaHussain (changes requested):

readme.md
kamaljohnson: Yeah I was going to do the same, but the naming which was used in the Dropdown component for this class is this 

https://github.com/frappe/frappe-ui/blob/b1dc1ff821682903d12068aca078902838719b4a/src/components/Dropdown.vue#L85-L88

I named the class in the same way to maintain consistency throughout.

I too think we should just use full-width throughout.

kamaljohnson: I have refactored the class tag from `button-full-width` -> `width-full`, and the refactor in the `Dropdown` component will be resolved with #2

PR #13, opened 2022-06-02
feat: ListManager component

kamaljohnson:
- [x] render lists based on options
 - [x] auto fetch item updates
 - [x] auto fetch new items using filters
 - [x] auto pagination
 - [x] list filtering function
 - [x] list sorting function
 - [x] item selection handle
    - [x] select each row item
    - [x] select-all
 - [x] item on click handle

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -198,39 +198,89 @@ Uses [`feather-icons`](https://github.com/feathericons/feather) under the hood.
 ```
 
 ### ListManager
+ListManager handles
+- filtering and sorting
+- pagination (auto/manual)
+- handle list item updates
+- options based style rendering
+- handle item click and selection events
+
+The component exposes `manager` which has the following methods: `loadMore` and `update` and also to access `selectedItems`
+
+`manager.loadMore()` loads more list items by calling the list api using the provided options
+
+`manager.update()` can be used to change the filters and sort_by options that are applied on the list
+```js
+manager.update({
+  filters: { status: 'Open' },
+  order_by: 'modified desc'
+})
+```
+
+option based rendering?
+class tags for each template can be passed via options, eg:
+```js
+options: {
+  style: {
+    header: 'space-x-2',
+    body: 'space-y-2',
+    row: {
+      base: 'space-x-2 hover:bg-gray-50',
+      selected: 'bg-blue-50 hover:bg-blud-100'
+    },
+    footer: ''
+  }
+}
+```
+
+Event handling
+- selection events should be explicityly overrided via `select` method in `<template #listItem="{ row, select }">`
+- if an item is selected then the on_click for the row is overrided for selection based events
+- a callback can be passed to `handle_row_click` options, which gets triggered when the row is clicked
+
+
 ```html
 <ListManager
   ref="list"
   :options="{
-    cache: ['Ticket', 'Desk'],
     doctype: 'Ticket',
     filters: {'status': 'Open'},
     fields: ['name', 'subject', 'status'],
     limit: 20,
     order_by: 'modified desc',
-    classes: {
-      body: 'space-y-2'
+    style: {
+      header: 'flex flex-row space-x-2',
+      body: 'space-y-2',
+      row: {
+        base: 'flex flex-row space-x-2',
+        selected: ''
+      }
     },
     auto_pagination: true,
-    handle_item_selection: true
+    handle_row_click: () => {
+      // do something...
+    }
   }"
   >
   <template #header="{ fields }">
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -21,6 +21,7 @@ export { default as Resource } from './components/Resource.vue'
 export { default as Spinner } from './components/Spinner.vue'
 export { default as SuccessMessage } from './components/SuccessMessage.vue'
 export { default as TextEditor } from './components/TextEditor'
+export { default as ListManager } from './components/ListManager'
```
Alchez, line 24: Can't this simply be:

```suggestion
export ListManager from './components/ListManager'
```

without the need to export `default` explicitly in the index file?

readme.md
```diff
@@ -198,39 +198,89 @@ Uses [`feather-icons`](https://github.com/feathericons/feather) under the hood.
...
+      type="checkbox" 
+      :checked="row.meta.selected" 
+      @click="select" 
+    />
+    <div>{{ row.data.field_1 }}</div>
+    <div>{{ row.data.field_2 }}</div>
+    ...
   </template>
   <template #footer>
+    <Button @click="() => { $refs.list.manager.loadMore() }">
```
netchampfaris, line 281: Better to expose `manager` via slots?

readme.md
```diff
@@ -198,39 +198,89 @@ Uses [`feather-icons`](https://github.com/feathericons/feather) under the hood.
...
+      header: 'flex flex-row space-x-2',
+      body: 'space-y-2',
+      row: {
+        base: 'flex flex-row space-x-2',
+        selected: ''
+      }
     },
     auto_pagination: true,
-    handle_item_selection: true
+    handle_row_click: () => {
```
netchampfaris, line 260: camelCasing ?

readme.md
```diff
@@ -198,39 +198,89 @@ Uses [`feather-icons`](https://github.com/feathericons/feather) under the hood.
...
     order_by: 'modified desc',
-    classes: {
-      body: 'space-y-2'
+    style: {
+      header: 'flex flex-row space-x-2',
+      body: 'space-y-2',
+      row: {
+        base: 'flex flex-row space-x-2',
+        selected: ''
+      }
```
netchampfaris, line 257: I think the List component should not render any HTML. It should be renderless.

src/components/ListManager/ListManager.vue
```diff
@@ -0,0 +1,212 @@
...
+    return {
+      manager,
+      newItems,
+      selectedItems,
+      selectionMode
+    }
+  },
+  mounted() {
+    this.manager.resource = this.$resources.list
+    this.handleRealtimeUpdate()
```
netchampfaris, line 139: Cleanup socket listener on unmount?

netchampfaris: Also, if you are using `setup` function, you can do the setup/cleanup of socket listener in there.

src/components/ListManager/ListManager.vue
```diff
@@ -0,0 +1,212 @@
...
+        }
+			}
+		},
+    handleRealtimeUpdate() {
+      this.$socket.on("list_update", (data) => {
+        if (data.doctype === this.options.doctype) {
+          this.$resources.document.fetch({
+            doctype: this.manager.options.doctype,
+            filters: {
+              ...this.options.filters,
```
netchampfaris, line 201: No need for any other filters if `name` filter is set

netchampfaris (changes requested):

src/index.js
kamaljohnson: @Alchez this is done just to make all the exports consistent

readme.md
kamaljohnson: > I think the List component should not render any HTML. It should be renderless.

I am using resource manager with type list within LIstManager component,
the style of the list can be fully customized by the user, by passing class tags to style options

since this component also handles auto pagination, item selection, click events, etc... I don't know how this could be handled without the list rendering

the renderless part could be handled using row resource manager too,

readme.md
kamaljohnson: @netchampfaris since I was using order_by inside options, which is directly passed to the list resource, to make all the option keys follow the same format.

src/components/ListManager/ListManager.vue
kamaljohnson: this was done to reduce unwanted API fetches, if the new updated item does not follow the filters passed via the user inside the list manager options, it does not make sense to handle the update either right

also, I think this is how the desk (framework) list UI handles list updates too

PR #25, opened 2022-08-26
fix: error when empty list passed to mentions prop in text editor

kamaljohnson:
Issue: When an empty list is passed to the `mentions` prop, it throws this error when `@` is typed to the text editor

https://user-images.githubusercontent.com/22856401/186871351-c1e950d8-c889-48f9-8d21-fe3bfce9025c.mov

Solution: checks and handles `mentions` block only if the list passed is not empty

### DIFF
--- a/src/components/TextEditor/mention.js
+++ b/src/components/TextEditor/mention.js
@@ -25,9 +25,12 @@ function getSuggestionOptions(options) {
     render: () => {
       let component
       let popup
+      let handleMentions = true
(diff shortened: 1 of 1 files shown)


netchampfaris (commented):
This kind of error handling was not needed. I have fixed this with https://github.com/frappe/frappe-ui/commit/ae9172619c18989b80660559cfd1c9d826b3e428

PR #46, opened 2023-02-21
fix: vite build errors with @tiptap/pm

dj12djdjs:
I was unable to build my project with vite.

as per: https://github.com/ueberdosis/tiptap/issues/3492#issuecomment-1419144848

issue is solved.


netchampfaris:
this is done via #53

PR #48, opened 2023-02-22
fix: correct endpoint methods with /api/method

dj12djdjs:
The requests are made to 
http://localhost:8000/frappe.client.get_list

This will always result it 404. 

Fixed by prefixing all requests with /api/method

### DIFF
--- a/src/resources/documentResource.js
+++ b/src/resources/documentResource.js
@@ -55,7 +55,7 @@ export function createDocumentResource(options, vm) {
     auto: options.auto || true,
     get: createResource(
       {
-        url: 'frappe.client.get',
+        url: '/api/method/frappe.client.get',
         makeParams() {
           return {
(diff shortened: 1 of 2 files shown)


NagariaHussain:
https://github.com/frappe/frappe-ui/blob/8769bd5982aec38f83efc511da6b051981afa05f/src/utils/frappeRequest.js#L22

The resourceFetcher automatically prepends `/api/method`

dj12djdjs:
@NagariaHussain It doesn't work. I could not find any usages of the `frappeRequest` export in `createResource`, `createListResource`, and `createDocumentResource`.

Therefor, all of the requests made when providing a method such as `frappe.client.get` will **not** call to `/api/method/frappe.client.get` but instead will be called to `/frappe.client.get` resulting in HTTP 404

PR #51, opened 2023-03-07
chore(deps): add @tiptap/pm as dependency and remove prosemirror-*

rutwikhdev:
Ran into some problems upgrading frappe-ui from older version to latest, related to this breaking change in tiptap depdency

Screenshot from 2023-03-07 19-50-26

Change ref: https://github.com/ueberdosis/tiptap/releases/tag/v2.0.0-beta.210

Installations that already have prosemirror or older version of tiptap won't have a problem but this will probably break in future I guess.

Was able to resolve for press by just ```yarn add @tiptap/pm``` but release guide suggested to remove the prosemirror-* too hence the change.


netchampfaris:
yarn.lock must also be updated, did it here: https://github.com/frappe/frappe-ui/pull/53

PR #55, opened 2023-03-16
fix(ux)!: white-listed method `onSuccess` argument must be `message`

ssiyad:
Example:
```typescript
whitelistedMethods:  {
  lastCommunication: {
    method: "last_communication",
    onSuccess: (arg) => {},
  },
}
````

Currently, `arg` is the parent doc, which we already have. It does make sense to return `data.message` instead, which is the return value of the white-listed method

### DIFF
--- a/src/resources/documentResource.js
+++ b/src/resources/documentResource.js
@@ -135,7 +135,7 @@ export function createDocumentResource(options, vm) {
               }
             }
           }
-          onSuccess?.call(vm, out.doc)
+          onSuccess?.call(vm, data.message)
         },
         ...otherOptions,
       },

### CODE REVIEW

File: src/resources/documentResource.js [hunk 1]
```diff
@@ -122,20 +122,7 @@ export function createDocumentResource(options, vm) {
...
-                doc.name.toString() === out.name.toString()
-              ) {
-                out.doc = transform(doc)
-                // update data in list resources
-                updateRowInListResource(out.doctype, out.doc)
-                break
-              }
-            }
-          }
-          onSuccess?.call(vm, out.doc)
```
netchampfaris, line 138: why remove this? this code updates the `out.doc` with the latest values from the server.

netchampfaris (changes requested):

src/resources/documentResource.js
ssiyad: Makes sense. But we can update all the docs in the list.

ssiyad:
```javascript
data.docs
  ?.forEach((doc) => updateRowInListResource(doc.doctype, transform(doc)))
  .filter((doc) => doc.doctype === out.doctype)
  .filter((doc) => doc.name.toString() === out.name.toString())
  .forEach((doc) => out.doc = transform(doc))
```

ssiyad:
```javascript
data.docs
  ?.map((doc) => {
    updateRowInListResource(doc.doctype, transform(doc));
    return doc;
  })
  .filter((doc) => doc.doctype === out.doctype)
  .filter((doc) => doc.name.toString() === out.name.toString())
  .forEach((doc) => out.doc = transform(doc))
```

netchampfaris: Yeah, but let's keep the change to minimum in this PR. We can do other changes in a different PR.

netchampfaris (approved):

PR #74, opened 2023-06-01
feat: `Autocomplete`: emit query change

ssiyad:


src/components/Autocomplete.vue
```diff
@@ -150,6 +150,11 @@ export default {
         .filter((group) => group.items.length > 0)
     },
   },
+  watch: {
+    query(q) {
+      this.$emit('queryChange', q)
```
netchampfaris, line 155: Use 'update:query' as event name to keep it consistent. Later we might expose query as a prop too

netchampfaris (approved):

PR #78, opened 2023-06-28
refactor: Separate file upload handler

surajshetty3416:
Move file upload handler class to a separate file so that it can be used directly in the script.


src/index.js
```diff
@@ -47,6 +47,7 @@ export { default as visibilityDirective } from './directives/visibility.js'
 export { default as call, createCall } from './utils/call.js'
 export { default as debounce } from './utils/debounce.ts'
 export { default as fileToBase64 } from './utils/file-to-base64.js'
+export { default as fileUploadHandler } from './utils/fileUploadHandler'
```
netchampfaris, line 50:
```suggestion
export { default as FileUploadHandler } from './utils/fileUploadHandler'
```

classes should be pascal case?

PR #52, opened 2023-03-08
feat: Espresso design system + Storybook

netchampfaris:
image

### Button.vue

`appearance` prop removed in favor of more flexible props for appearance: `variant` and `theme`

| Before | After |
|---|---|
|`<Button appearance="primary">`|`<Button variant="solid">`|
|`<Button appearance="danger">`|`<Button variant="solid" theme="red">`|
|`<Button appearance="minimal">`|`<Button variant="ghost">`|
|`<Button appearance="white">`|`<Button variant="outline">`|
|`<Button icon-left="plus">`|`<Button><template #prefix><FeatherIcon name="plus" class="w-4" /></template></Button`|
|`<Button icon-right="plus">`|`<Button><template #suffix><FeatherIcon name="plus" class="w-4" /></template></Button`|

### Dialog.vue
| Before | After |
|---|---|
|`<Dialog :options="{ actions: [label: 'Save', handler: fn ] }">`|`<Dialog :options="{ actions: [label: 'Save', onClick: fn ] }">`|

### Dropdown
| Before | After |
|---|---|
|`<Dropdown :items="[label: 'Save', handler: fn ]">`|`<Dropdown :items="[label: 'Save', onClick: fn ]">`|

### vite config

```diff
import { defineConfig } from 'vite'
import vue from '@vitejs/plugin-vue'
+ import frappeui from 'frappe-ui/vite'
- import { getProxyOptions } from 'frappe-ui/src/utils/vite-dev-server'

export default defineConfig({
  plugins: [
+  frappeui(),
    vue(),
    ...
  ],
- server: {
-   port: 8080,
-   proxy: getProxyOptions({ port: webserver_port }),
- },
  build: {
    ...
  },
})

```

PR #81, opened 2023-07-19
fix: Update `vite-dev-server.js` Node IP resolution fix

uhrjun:
Node more correctly follows OS native [`dns.lookup()`](https://github.com/nodejs/node/pull/39987) resulting in this breakage


shariquerik:
Fixed https://github.com/frappe/frappe-ui/pull/52/commits/232b17f95889abe5df7bf8e37553f1ecf51de041

PR #92, opened 2023-08-11
fix(switch): useless emit when value is `false`

ssiyad:


src/components/Switch.vue
```diff
@@ -54,15 +59,7 @@ const props = withDefaults(defineProps<SwitchProps>(), {
...
-
-// Handle the enabled state
-const enabled = ref(true)
-onMounted(() => {
-  enabled.value = props.modelValue as boolean
-})
-watch(enabled, (value) => {
-  emit('update:modelValue', value)
-})
+const modelValue = toRef(props, 'modelValue')
```
netchampfaris, line 62: why this and not `props.modelValue`?

ssiyad: to keep reactivity

PR #98, opened 2023-08-26
fix: allow html in tooltip

shariquerik:
image

### DIFF
--- a/src/components/Tooltip.vue
+++ b/src/components/Tooltip.vue
@@ -13,6 +13,11 @@
             {{ text }}
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


src/components/Tooltip.vue
```diff
@@ -13,6 +13,11 @@
             {{ text }}
           </div>
         </div>
+        <div
+          v-else-if="html"
```
netchampfaris, line 17: Better to use the `body` slot instead of passing an html string

netchampfaris (changes requested):

PR #102, opened 2023-09-25
feat: Tabs component

shariquerik:
```
<Tabs
	v-model="tabIndex"
	v-slot="{ tab }"
	:tabs="[
		{ label: 'Tab 1', content: 'Content of tab 1'},
		{ label: 'Tab 2', content: 'Content of tab 2' }
	]"
>
  <div class="p-5">{{ tab.content }}</div>
</Tabs>
```
**Default:**
By setting `tabIndex` you can set the default tab selected

**Change Tab:**
By changing the same `tabIndex` you can also change the tab

https://github.com/frappe/frappe-ui/assets/30859809/99620315-d50d-43c6-a738-2b1df18d2c39

**Tab with Icon:**
Here icon is a component.
```
const tabs = [
	{
		label: 'Activity',
		icon: ActivityIcon
	},
	{
		label: 'Emails',
		icon: EmailIcon
	},
	{
		label: 'Calls',
		icon: CallIcon
	}
]
```
image

**Need a custom tab button UI use tab slot**
```
<Tabs>
	<template #tab="{ tab, selected }">
		// custom tab button
	</template>
</Tabs>
```

**FrappeUI Storybook**
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@tiptap/starter-kit": "^2.0.3",
     "@tiptap/suggestion": "^2.0.3",
     "@tiptap/vue-3": "^2.0.3",
+    "@vueuse/core": "^10.4.1",
     "feather-icons": "^4.28.0",
     "idb-keyval": "^6.2.0",
     "showdown": "^2.1.0",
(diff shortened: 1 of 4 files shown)


netchampfaris:
@shariquerik Does this render content for all tabs at once? We should probably find a way to only render tab content for the current tab.

shariquerik:
> @shariquerik Does this render content for all tabs at once? We should probably find a way to only render tab content for the current tab.

It does render content for current tab

PR #111, opened 2023-10-30
fix: file uploader: consider `doctype` & `docname` separately

ssiyad:
In `FileUploader`, `doctype` is ignored if `docname` is not present. This PR change the behaviour to consider `doctype`, `docname` and `fieldname` as separate, unrelated properties. Please let me know if there is a reason to couple `doctype` and `docname` together.

### DIFF
--- a/src/utils/fileUploadHandler.ts
+++ b/src/utils/fileUploadHandler.ts
@@ -109,12 +109,16 @@ class FileUploadHandler {
         form_data.append('file_url', options.file_url)
       }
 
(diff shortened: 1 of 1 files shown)


netchampfaris:
@ssiyad what is the case when `doctype` has to be set without `docname`?

ssiyad:
It useful when the `doc` is not yet created but you still have to upload files for it. By referencing a doctype, you can technically `claim` it for your app. Which is not ideal but I think it might be slightly better than current implementation

PR #117, opened 2023-12-01
feat(Autocomplete): add support for disabled attribute

krantheman:
Autocomplete cannot be disabled currently :(


src/components/Autocomplete.vue
```diff
@@ -120,7 +121,7 @@ import FeatherIcon from './FeatherIcon.vue'
 
 export default {
   name: 'Autocomplete',
-  props: ['modelValue', 'options', 'placeholder'],
+  props: ['modelValue', 'options', 'placeholder', 'isDisabled'],
```
netchampfaris, line 124: prop name should be `disabled` consistent with Input, Select, Textarea, etc.

src/components/Autocomplete.vue
```diff
@@ -7,6 +7,7 @@
             <button
               class="flex h-7 w-full items-center justify-between gap-2 rounded bg-gray-100 px-2 py-1 transition-colors hover:bg-gray-200 focus:ring-2 focus:ring-gray-400"
               :class="{ 'bg-gray-200': isComboboxOpen }"
```
this-gavagai, line 9: There might also be a conditional class application here to indicate that the line is disabled, such as replacing the current line with:
```
:class="[ {'bg-gray-200': isComboboxOpen}, disabled ? 'text-gray-600 bg-gray-50 placeholder-gray-400 hover:bg-gray-50' : '' ]"
```

PR #135, opened 2024-03-23
feat: add pages to vite plugin for builder

blaggacao:
- required by https://github.com/frappe/builder/blob/95493c56ce9e98cadca45c226c3efb51e781c36e/frontend/vite.config.js#L34

cc @surajshetty3416

### DIFF
--- a/vite.js
+++ b/vite.js
@@ -1,27 +1,30 @@
 const path = require('path')
 const fs = require('fs')
 
-module.exports = function proxyOptions({ port = 8080 } = {}) {
+module.exports = function proxyOptions({
+  port = 8080,
+  regex = '^/(app|login|api|assets|files)',
+ } = {}) {
   const config = getCommonSiteConfig()
   const webserver_port = config ? config.webserver_port : 8000
   if (!config) {
     console.log('No common_site_config.json found, using default port 8000')
+  }
+  let proxy = {}
+  proxy[regex] = {
+    target: `http://127.0.0.1:${webserver_port}`,
+    ws: true,
+    router: function (req) {
+      const site_name = req.headers.host.split(':')[0]
+      return `http://${site_name}:${webserver_port}`
+    }
   }
   return {
     name: 'frappeui-vite-plugin',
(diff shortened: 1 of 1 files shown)


surajshetty3416:
@blaggacao Why is this required here 🤔

blaggacao:
Ah, you're right. That's a better idea! Will do, shortly.

blaggacao:
@surajshetty3416 Does this look better?

Btw, I don't know why not this diff, but I left it how it was

```diff
module.exports = function proxyOptions({
  port = 8080,
  regex = '^/(app|login|api|assets|files)',
- } = {}) {
+ }) {
```

surajshetty3416:
> Btw, I don't know why not this diff, but I left it how it was

the `{}` in it is just a default value so that you can call `proxyOptions()` (without any parameter) and still be able to destructure the object with default values (eg. port and regex).

vite.js
```diff
@@ -1,27 +1,30 @@
 const path = require('path')
 const fs = require('fs')
 
-module.exports = function proxyOptions({ port = 8080 } = {}) {
+module.exports = function proxyOptions({
+  port = 8080,
+  regex = '^/(app|login|api|assets|files)',
```
surajshetty3416, line 6:
```suggestion
  source = '^/(app|login|api|assets|files)',
```
Variable name "source" sounds more apt? 
like "all these sources are redirected to a different target"

vite.js
```diff
@@ -1,27 +1,30 @@
...
+ } = {}) {
   const config = getCommonSiteConfig()
   const webserver_port = config ? config.webserver_port : 8000
   if (!config) {
     console.log('No common_site_config.json found, using default port 8000')
   }
+  let proxy = {}
+  proxy[regex] = {
```
surajshetty3416, line 14:
```suggestion
  proxy[source] = {
```

surajshetty3416 (changes requested):

surajshetty3416 (approved):

PR #140, opened 2024-04-19
fix: also allow changing file type change while opening file selector

shariquerik:


src/components/FileUploader.vue
```diff
@@ -51,7 +51,11 @@ export default {
     },
   },
   methods: {
-    openFileSelector() {
+    openFileSelector(type) {
+      if (type) {
+        // change inputs filetype to accept the type of file you want to upload
+        this.$refs['input'].accept = type
```
netchampfaris, line 57: There is a fileTypes prop as well, this will make the prop value and input attribute out of sync.

PR #170, opened 2024-08-12
Pass site_config as an optional arg.

bprashanth:
This change passes a path to a `site_config` into `vite.js`'s `proxyOptions` so callers that don't have a standard `sites/` layout (eg frappe press dashboard - potentially any caller using `frappe-ui` as a lib) can configure the `webserver_port`.

Also includes basic node test run via `node --test vite.test.js`.


surajshetty3416:
@bprashanth please resolve conflicts and reopen if needed

PR #173, opened 2024-08-16
feat: Tree Component

GursheenK:
#### Description
Added a Tree Component which takes an object containing nested data to be rendered as a tree. 

#### Props 
1. **node** - root node ie. complete tree data where each node consists of the label and and it's children which are then used to recursively render the sub-trees. 

2. **options** - Options to change the `rowHeight`, `indentWidth` and `showLines` properties of the Tree.

#### Slots
Added slots to override the label, icon or the complete node. Documented the slot usage and all other details in the Storybook docs.

<br>

<details>

<summary> Screen Recording </summary>

<br>

https://github.com/user-attachments/assets/1cc5f169-cc8b-4624-8023-f2db4b45d3c3

</details>


src/components/Tree/Tree.vue
```diff
@@ -0,0 +1,136 @@
...
+const props = withDefaults(
+  defineProps<{
+    node: TreeNode
+    options?: TreeOptions
+  }>(),
+  {
+    options: () => ({
+      rowHeight: '25px',
+      indentWidth: '20px',
+      showLines: true,
```
ruchamahabal, line 95: should we call this `showLevelMarkers` instead of showLines?

src/components/types/Tree.ts
```diff
@@ -0,0 +1,10 @@
+export type TreeNode = {
+  label: string
```
ruchamahabal, line 2: Should we also have a `nodeKey` in tree (like `rowKey` in list) since titles of 2 nodes could be same

ruchamahabal (changes requested):

ruchamahabal:
> . Although it is functional, unsure where to add this to maintain a clear separation between Frappe-specific and other components.

We can have a separate folder "frappe" under components for frappe specific components

surajshetty3416 (approved):
LGTM

PR #177, opened 2024-08-22
fix: increase Tooltip's z-index to avoid overlap

ruchamahabal:
Currently, the tooltip z-index is 10. Gets hidden by other layers having z-index. Popover has a z-index of 100: https://github.com/frappe/frappe-ui/blob/96ecd6adf3669bd231e59ee442ac9fb3711a840d/src/components/Popover.vue#L20
Bumping it up for Tooltip too

image

**After**

image


NagariaHussain:
https://fantinel.dev/css-isolation-property

ruchamahabal:
> [fantinel.dev/css-isolation-property](https://fantinel.dev/css-isolation-property)

Yup, this helps too! But 10 is very low, so bumped it up to be consistent with Popover.

PR #195, opened 2024-10-16
feat: Init dark mode

surajshetty3416:
- Added Espresso colors and variables. [Espresso Figma](https://www.figma.com/community/file/1407648399328528443)
	Screenshot 2024-10-16 at 11 04 48 AM
- Using `data-theme='dark'` as selector instead of `dark` class as it looks more semantic and might rarely create conflict.

#### Usage:
- Example components are from Espresso Figma file.
- Setting variable class will work for both light and dark mode.
- In your app, you'll have to create a toggler that toggles the `data-theme` attribute on root from "light" to "dark"

**Example 1**: 
So now for following style we would use `bg-surface-blue-2` class instead of `bg-blue-500` 
Screenshot 2024-10-16 at 11 06 34 AM

**Example 2:**
use `border-outline-green-2` class instead of `border-green-500` 
Screenshot 2024-10-16 at 11 12 59 AM

moved from [Builder](https://github.com/frappe/builder)


surajshetty3416:
[Refactored Badge component](https://github.com/frappe/frappe-ui/pull/195/commits/625f8eb80d913515cdad639fc61b1eb830d6b39f) as an example to make it DM compatible. 
I'll merge this for now so that others can start using the espresso variables in new components.

We'll also have to refactor existing components one-by-one.

PR #207, opened 2024-10-23
fix(pageMeta): Don't access `document` in top-level

cogk:
Improve compatibility with tools that set the `document` global to undefined while pre-rendering/SSR (e.g. Nitro).

https://stackoverflow.com/questions/46058544/document-is-not-defined-in-nuxt-js


netchampfaris:
@cogk

Hey are you able to use `frappe-ui` in a nuxt project with this change?

PR #217, opened 2024-11-16
fix: dialog action context object

netchampfaris:
- earlier context object was passed, but mistakenly got changed into close function being passed directly
- fix types

#### Before
```
<Dialog :actions="[
 {
  label: 'Submit',
  onClick(close) {
    close() // close dialog
  }
 }
]">
```

#### After
```
<Dialog :actions="[
 {
  label: 'Submit',
  onClick({ close }) {
    close() // close dialog
  }
 }
]">
```

### DIFF
--- a/src/components/Dialog.vue
+++ b/src/components/Dialog.vue
@@ -116,13 +116,13 @@
               </slot>
               <div
                 class="px-4 pb-7 pt-4 sm:px-6"
-                v-if="dialogActions.length || $slots.actions"
+                v-if="actions.length || $slots.actions"
               >
                 <slot name="actions" v-bind="{ close }">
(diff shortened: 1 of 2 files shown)


src/components/Dialog.vue
```diff
@@ -209,15 +212,16 @@ watch(
       let loading = ref(false)
       return {
         ...action,
-        loading,
+        loading: loading.value,
```
ruchamahabal, line 215: This breaks reactivity IIRC. Even I had tried changing it to be mapped to value

netchampfaris: @ruchamahabal Thanks for pointing out, fixed.

PR #228, opened 2024-11-28
fix: add onClick with route in breadcrumb

RitvikSardana:
**Issue:**
image

Currently when the we add a route along with onClick property to a "breadcrumb", the route takes the precedence, as shown below.
image
This results in a problem where we want to trigger a function while clicking on a breadcrumb.

But if we just add onClick and remove the "route" property like this
image

Then the breadcrumb results in a button
image

which is a bad UI.

**Result**
1.  API
image

2.  Both focus and hover classes has been made inherit
image

### DIFF
--- a/src/components/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs.vue
@@ -60,11 +60,10 @@
 </template>
 <script setup lang="ts">
 import { useWindowSize } from '@vueuse/core'
-import { computed } from 'vue'
+import { computed, h } from 'vue'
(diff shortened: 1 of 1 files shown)


netchampfaris:
This is not the correct fix. The idea is to render a normal button and not frappe-ui Button instead of removing styles from Button explicitly.

Fixed here: https://github.com/frappe/frappe-ui/commit/c89f7068ddbfdebe1f29456b1992fb0f6f87aba3

PR #234, opened 2024-12-03
fix: Replaced luxon with dayjs

shariquerik:
Old PR: https://github.com/frappe/frappe-ui/pull/229

`dayjs` & `dayjsLocal` is now accessible via frappe-ui

Examples:
```js
import { dayjs, dayjsLocal } from 'frappe-ui'

// system timezone => "Asia/Kolkata"
// local(user) timezone => "Asia/Tokyo"

let date = '2024-12-03 19:06:46'

// without timezone
date = dayjs(date).format('DD/MM/YYYY HH:mm:ss')
// result => 03/12/2024 19:06:46

// with timezone
date = dayjsLocal(date).format('DD/MM/YYYY HH:mm:ss')
// result => 03/12/2024 22:36:46
```
> check [dayjs docs](https://day.js.org/docs/en/installation/installation) for more info

How to set system & user timezone's?

Timezone is coming from boot which makes it available in window. You can set config in `App.vue` since window.timezone is not accessible in `main.js`. (Faris is working on a better solution for this, also will have to standardise where to save these info since saving in window is not a good practise ig)
```js
import { setConfig } from 'frappe-ui'

setConfig('systemTimezone', window.timezone?.system)
setConfig('localTimezone', window.timezone?.user)
```


src/utils/dates.ts
```diff
@@ -1,13 +1,4 @@
...
-
-declare global {
-  interface Window {
-    timezone?: {
-      system?: string
-      user?: string
-    }
-  }
-}
+import { dayjs } from './dayjs'
```
netchampfaris, line 1: why is this file needed?

shariquerik: moved all datepicker components to separate folder with dates.ts -> utils.ts

PR #252, opened 2025-01-15
fix: 1.15 lineheight is too small

shariquerik:
Changed line-height from 1.15 -> 1.4 for (2xs to xl), 1.3 for 2xl & 1.25 for 3xl

| Size | Before | After |
|-----|--------|------|
| 2xs |image|image|
| xs |image|image|
|  sm |image|image|
|base|image|image|
| lg |image|image|
| xl |image|image|
| 2xl |image|image|
| 3xl |image|image|

### DIFF
--- a/src/tailwind/plugin.js
+++ b/src/tailwind/plugin.js
@@ -87,7 +87,7 @@ module.exports = plugin(
         '2xs': [
           '11px',
           {
-            lineHeight: '1.15',
+            lineHeight: '1.4',
             letterSpacing: '0.01em',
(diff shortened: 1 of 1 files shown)


netchampfaris:
If you want larger line heights, use `text-p-base`, `text-p-sm` classes instead of `text-base`, `text-sm`. 

Since this is a common issue, maybe we should switch the default behaviour and have larger line heights in `text-base`, `text-sm` classes and have tighter line heights in `text-tight-base`, `text-tight-sm`, etc.

Issue #263, opened 2025-02-14
Autocomplete: Issue in multiple selection


kakde-sandeep:
On clicking on same option multiple times instead of selecting & deselecting, option is added to the selected list.

![Image](https://github.com/user-attachments/assets/3c4e6d9c-9098-4975-97c9-ff327878d500)

PR #265, opened 2025-02-18
fix: Autocomplete multiselect

kakde-sandeep:
Combox's default comparison is not working as expected, so providing users to override this comparison according to there needs.
In this commit I've updated the code for autocomplete story with Multiple options & Search to provide the example of overriding the combox's **by**  function.

Fixes: https://github.com/frappe/frappe-ui/issues/263

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frappe-ui",
-  "version": "0.1.166",
+  "version": "0.1.167",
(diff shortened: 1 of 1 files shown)


kakde-sandeep:
@safwansamsudeen Can you please check this

kakde-sandeep:
@netchampfaris should we add a default function as well for this instead of user explicitly passing the comparison function. What are your thoughts on this?

PR #271, opened 2025-03-12
feat: enhancements based on drive

safwansamsudeen:
### List view
1. As opposed to this less clean look...
image

... now rows are grouped together when selected consecutively:
image

2. Tooltips are centered onto their subject label, not the entire row item:
image

3. Active (the last clicked) support has been added - configure with `enableActive`, defaulting to `false.`
4. Tooltip text can be configured now by an individual column `getTooltip` function, like the label. If it's not, it defaults to the relevant `key` in the row, _not_ the column label through `getLabel`. It's not intuitive, but is the current behavior, so I'm maintaining it.

### Others
1. The text  in the floating action bar can be changed with a list option `selectionText`.
2. There's some proximity detection in the checkbox - clicking close to it doesn't trigger row click actions
3. In breadcrumbs, `onClick` and `rowRoute` are not mutually exclusive anymore
4. The loading indicator can be scaled


safwansamsudeen:
Changes: 
- the original floating action bar behavior has been retained.
- **Tooltips are rendered in HTML** instead of plain text (used in Drive)
- Bugs in row selection in the case of groups

src/components/ListView/ListRow.vue
```diff
@@ -112,4 +119,20 @@ const rowHeight = computed(() => {
...
+  let groups = list.value.rows[0]?.group
+    ? list.value.rows.map((k) => k.rows)
+    : [list.value.rows]
+
+  for (let rows of groups) {
+    let currentIndex = rows.findIndex((k) => k == props.row)
+    if (currentIndex === -1) continue
+    let atBottom = !selections.includes(rows[currentIndex + 1]?.name)
+    let atTop = !selections.includes(rows[currentIndex - 1]?.name)
+    return (atBottom ? 'rounded-b ' : '') + (atTop ? 'rounded-t' : '')
```
netchampfaris, line 135: return from for loop?

src/components/Tooltip/Tooltip.vue
```diff
@@ -53,7 +53,7 @@ const delayDuration = computed(() => props.hoverDelay * 1000)
             <div
               class="rounded bg-surface-gray-7 px-2 py-1 text-xs text-ink-white shadow-xl"
             >
-              <div>{{ props.text }}</div>
+              <div v-html="props.text"></div>
```
netchampfaris, line 56: use slot if you want html content? doesn't make sense to pass html content via `text` prop

src/components/ListView/ListView.vue
```diff
@@ -81,6 +78,7 @@ let _options = computed(() => {
     getRowRoute: props.options.getRowRoute || null,
     onRowClick: props.options.onRowClick || null,
     showTooltip: defaultTrue(props.options.showTooltip),
+    selectionWord: props.options.selectionWord || ((val) => val === 1 ? 'Row' : 'Rows'),
```
netchampfaris, line 81: this is not translation friendly, instead it should be something like:
```js
selectionText: props.options.selectionText || val => val === 1 ? '1 row selected' : '${val} rows selected'
```

src/components/ListView/ListRowItem.vue
```diff
@@ -1,27 +1,27 @@
...
+        :is="
+          typeof column.prefix === 'function'
+            ? column.prefix({ row })
+            : column.prefix
+        "
+      />
+    </slot>
+    <slot v-bind="{ label }">
+      <component
+        :is="list.options.showTooltip ? Tooltip : 'div'"
```
netchampfaris, line 15: it should not render extra `div` if Tooltip is not rendered

netchampfaris (changes requested):

src/components/ListView/ListRow.vue
safwansamsudeen: Yes - `currentIndex` (the index of the row) is only found in one group.

src/components/Tooltip/Tooltip.vue
safwansamsudeen: Changed.

src/components/ListView/ListRowItem.vue
safwansamsudeen: This is the label - it renders a tooltip along with the label text if the tooltip should be rendered; and if not, it displays the label alone.

PR #272, opened 2025-03-14
feat: add `label`, `loading`, `placement` & `show-footer` props

nextchamp-saqib:
1. Add `label` prop similar to `FormControl`
2. Add `loading` prop to display loading indicator
3. Add `show-footer` prop to show `clear` button even in case of `single` value

CleanShot 2025-03-14 at 15 16 06@2x

4. minor change in rating styles
  - Before:
     CleanShot 2025-03-14 at 16 06 21@2x

  - After
    CleanShot 2025-03-14 at 16 05 45@2x

### DIFF
--- a/src/components/Autocomplete.vue
+++ b/src/components/Autocomplete.vue
@@ -318,13 +318,17 @@ const filterOptions = (options: Option[]) => {
 const selectedValue = computed({
   get() {
     if (!props.multiple) {
-      return findOption(props.modelValue as AutocompleteOption)
+      return (
+        findOption(props.modelValue as AutocompleteOption) ||
+        // if the modelValue is not found in the option list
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete.vue
```diff
@@ -60,12 +79,17 @@
...
                   />
-                  <button
+                  <div
                     class="absolute right-0 inline-flex h-7 w-7 items-center justify-center"
-                    @click="clearAll"
                   >
-                    <FeatherIcon name="x" class="w-4 text-ink-gray-8" />
-                  </button>
+                    <LoadingIndicator
+                      v-if="!props.loading"
```
netchampfaris, line 86: @nextchamp-saqib bug?

PR #294, opened 2025-04-17
fix: button label overflow

nextchamp-saqib:
Before:
CleanShot 2025-04-17 at 12 36 20@2x

After:
CleanShot 2025-04-17 at 12 35 34@2x


surajshetty3416:
Ideally, the button label should be short enough that it is completely visible.

--

If someone wants to read full label in a truncated label button, we can have "title" attribute set so that full text is visible on hover. But this might only work on desktop.

--

Merging this for now since the fix makes button look acceptable in those edge cases. IMO It's up to developer to use shorter labels or adjust widths.

surajshetty3416 (approved):

PR #297, opened 2025-04-26
feat: added find and replace functionality to texteditor

shahzeelahmed:
### Issue
Navigating through long documents to find a word can be painful. this PR introduces find and replace capability in the text editor component.
### features
* higlights each letter in the editor as its typed
* replace currently highlighted word or all occurrences 
* unselect searched text on popover close
* case matching
* total number of occurrences of searched text
### demo

https://github.com/user-attachments/assets/ab86572b-4209-4cfc-9ade-eeeb00cd3c01


rmehta:
Why are these old PRs still open?

PR #312, opened 2025-05-23
added lable in DatePicker

NihalRoshanCK:
Added support for labels to the date picker component. Previously, when a date was selected, the placeholder would disappear, leaving the field context unclear. This update ensures that a label is always visible when props.label is provided, maintaining clarity even after a value is selected.


NihalRoshanCK:
image

PR #313, opened 2025-05-23
Add unique in  Autocomplete to handle duplicates

NihalRoshanCK:
Added unique prop to Autocomplete field to prevent duplicate selections.
This update introduces a unique boolean prop to the Autocomplete component. When enabled, it ensures that the same item cannot be selected multiple times, improving data integrity in multi-select scenarios.

### DIFF
--- a/src/components/Autocomplete.story.vue
+++ b/src/components/Autocomplete.story.vue
@@ -99,5 +99,17 @@ const options = [
         />
       </div>
     </Variant>
+    <Variant title="Multiple options without duplicates">
+      <div class="p-2">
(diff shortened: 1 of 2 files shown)


NihalRoshanCK:
image

nextchamp-saqib:
https://github.com/frappe/frappe-ui/issues/320#issuecomment-2939802155

NihalRoshanCK:
https://github.com/frappe/frappe-ui/pull/355 also resolves the issue but it include compareFn in type which i thing not needed

PR #323, opened 2025-06-03
feat: ability for alternative auth in useList & useDoc

### LINKED ISSUE #322: feat: useList requires headers & credentials fetchOptions for cross-origin requests
UseListOptions allows a `baseUrl` but does not allow the headers and credentials fetch options to be passed down to the fetch call.

Use-case is cross-origin authenticated requests from frappe-ui apps (mainly for local development using a cloud frappe instance as backend)

https://github.com/frappe/frappe-ui/blob/8313663f3f92a81da2c927d63cfd6d27c8e61c67/src/data-fetching/useList/types.ts#L28

casesolved-co-uk:
Ability for alternative auth in useList & useDoc

Use-case is local UI development using frappe instance as backend.

Requires server set up of allow_cors and application to provide auth headers

fixes #322


netchampfaris:
@casesolved-co-uk Thanks for your PR. Can you add a test for this?

netchampfaris:
Thanks @casesolved-co-uk! We've redone this as a smaller, minimal-diff version in #806 (configurable `requestBaseUrl` + `requestHeaders`, with tests), covering the same use case from #322. Closing this in favor of that.

PR #327, opened 2025-06-12
fix: improve dropdown

safwansamsudeen:


safwansamsudeen:
And allowing index in the breadcrumbs (used to check if it's the first breadcrumb and add an icon).

src/components/Dropdown.vue
```diff
@@ -40,26 +40,35 @@
                 :is="item.component"
                 :active="active"
               />
+              <div v-else-if="item.divider">
+                <hr class="m-1 bg-gray-300" />
+              </div>
```
netchampfaris, line 45: m-1 applies margin on all sides? you probably want margin only on y-axis. Check Figma design. Also, use semantic class for bg, `bg-surface-gray-*` instead of `bg-gray-`

src/components/Dropdown.vue
```diff
@@ -40,26 +40,35 @@
...
                 <FeatherIcon
                   v-if="item.icon && typeof item.icon === 'string'"
+                  :class="item.color"
                   :name="item.icon"
                   class="mr-2 h-4 w-4 flex-shrink-0 text-ink-gray-6"
                   aria-hidden="true"
                 />
                 <component
                   class="mr-2 h-4 w-4 flex-shrink-0 text-ink-gray-6"
+                  :class="item.color"
```
netchampfaris, line 64: color? you mean class?

safwansamsudeen:
Will be done in #370 instead.

PR #328, opened 2025-06-12
feat: Add "Text Justify" in the TextEditor component

### LINKED ISSUE #284: Add "Text Justify" in the TextEditor component
Right now the text editor supports only Left, right and center aligned. It would be great if the "Justify" is also added in the text editor options.

radeqq007:
Resolves #284

### DIFF
--- a/docs/components/text-editor.md
+++ b/docs/components/text-editor.md
@@ -32,6 +32,7 @@ let fixedMenuButtons = [
     'Align Left',
     'Align Center',
     'Align Right',
+    'Align Justify',
     'FontColor',
     'Separator',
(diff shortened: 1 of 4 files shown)


netchampfaris:
Text alignment (including justify) is already supported in the newer Editor molecule (`src/molecules/editor`) via the `TextAlign` extension, e.g. `editor.chain().setTextAlign('justify')`. We're intentionally keeping justify out of the default toolbar so apps can opt in through their own menu config.

This PR targets the older `TextEditor` component, which we're phasing out in favor of the Editor molecule, so I'm going to close it.

PR #339, opened 2025-06-23
feat: improved tab buttons

safwansamsudeen:
Aligns with Espresso.

Before (left-most and right-most components): height is different from other components, border radius is inconsistent, disabling isn't possible, and only feather icon is accepted.
image

After:
image


nextchamp-saqib:
It looks like this after changes:

CleanShot 2025-07-09 at 16 11 17@2x

PR #345, opened 2025-06-26
feat: enable auto import of icons

safwansamsudeen:
Currently, icons are automatically available in templates, but not in the `script`. This leads to code like [this](https://github.com/frappe/drive/blob/b0d9f92826c83a69740d4f9ce254ed6b278010be/frontend/src/components/Navbar.vue#L133).

For both ease of access and consistency, this PR enables auto-importing of icons in the script. You can now simply use `LucideSearch` without needing to import it explicitly.

```vue
<script setup lang="ts">
const button = {
  label: 'Search',
  icon: LucideSearch,
}
</script>
```

Update your `tsconfig.json` to include the new `auto-imports.d.ts` file, which will be generated by the `unplugin-auto-import` package. This file will contain type definitions for all auto-imported icons, ensuring that TypeScript recognizes them.

```
{
  "include": [..., "./auto-imports.d.ts", "./components.d.ts"],
}
```

_Side note_: users often import Lucide icons from `lucide-vue-next`, as that is what autocomplete suggests. But icons from there aren't part of frappe-ui and thus aren't Espresso aligned. This PR will also fix for that trap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "socket.io-client": "^4.5.1",
     "tippy.js": "^6.3.7",
     "typescript": "^5.0.2",
+    "unplugin-auto-import": "^19.3.0",
     "unplugin-icons": "^22.1.0",
     "unplugin-vue-components": "^28.4.1"
   },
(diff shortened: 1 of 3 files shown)


netchampfaris:
@safwansamsudeen 

> _Side note_: users often import Lucide icons from `lucide-vue-next`, as that is what autocomplete suggests. But icons from there aren't part of frappe-ui and thus aren't Espresso aligned. This PR will also fix for that trap.



How is this fixed?

PR #351, opened 2025-07-02
fix: remove implict overwrite of response data

nextchamp-saqib:


src/data-fetching/useFrappeFetch.ts
```diff
@@ -27,7 +27,11 @@ export const useFrappeFetch = createFetch({
         docStore.setDocs(docs)
         listStore.updateRows(docs)
       }
-      ctx.data = responseData.data
+      if (responseData.message) {
+        ctx.data = responseData.message
+      } else if (responseData.data) {
+        ctx.data = responseData.data
+      }
```
netchampfaris, line 34: useFrappeFetch is only meant for `/api/v2`

PR #353, opened 2025-07-02
fix: image pasting via clipboard

RitvikSardana:
**Issue:**
When we paste images via clipboard into the text editor we get the following result:
telegram-cloud-photo-size-5-6278269630518773974-y

**Reason:**
We only handle Text pasting from clipboard
ref:
telegram-cloud-photo-size-5-6280429891694479811-y

**Fix:**
Find images from clipboard and use `processMultipleImages` function to upload images via clipboard

1. Image drop works like before
2. Image pasting via clipboard also works

https://github.com/user-attachments/assets/8d583d01-f095-4b86-8ac0-9060b48bc685

### DIFF
--- a/src/components/TextEditor/extensions/markdown-paste-extension.ts
+++ b/src/components/TextEditor/extensions/markdown-paste-extension.ts
@@ -2,6 +2,8 @@ import { Extension } from '@tiptap/core'
 import { Plugin, PluginKey } from '@tiptap/pm/state'
 import { DOMParser } from '@tiptap/pm/model'
 import { detectMarkdown, markdownToHTML } from '../../../utils/markdown'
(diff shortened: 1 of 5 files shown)


src/components/TextEditor/extensions/markdown-paste-extension.ts
```diff
@@ -2,6 +2,8 @@ import { Extension } from '@tiptap/core'
 import { Plugin, PluginKey } from '@tiptap/pm/state'
```
netchampfaris, line 2: this extension was originally supposed to handle markdown paste only. now that you've added image processing you should rename the extension name.

netchampfaris (changes requested):

PR #355, opened 2025-07-05
feat: Adding default function for autocomplete option comparison

kakde-sandeep:
As per discussion on https://github.com/frappe/frappe-ui/pull/265, I've added the default function for comparison.
@netchampfaris please check now.

Fixes: #263

### DIFF
--- a/src/components/Autocomplete/types.ts
+++ b/src/components/Autocomplete/types.ts
@@ -26,6 +26,7 @@ export type AutocompleteProps = {
   loading?: boolean
   placement?: string
   showFooter?: boolean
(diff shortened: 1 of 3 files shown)


src/components/Autocomplete/types.ts
```diff
@@ -26,6 +26,7 @@ export type AutocompleteProps = {
   loading?: boolean
   placement?: string
   showFooter?: boolean
+  by?: (a: Option, b: Option) => boolean
```
nextchamp-saqib, line 29:
```suggestion
  compareFn?: (a: Option, b: Option) => boolean
```

PR #356, opened 2025-07-05
Feat(datepicker): add minDate and maxDate props to restrict selectable dates

LeoDanielA01:
This PR introduces minDate and maxDate props to the DatePicker component, allowing users to restrict which dates can be selected from the calendar.


ishanloya:
This would be very useful. TimePicker already has minTime and maxTime so this would make them consistent with each other.

netchampfaris:
@LeoDanielA01 All datepicker components now have `min` and `max` prop.

PR #377, opened 2025-08-11
fix: Autocomplete component  Disabled State

suhail135:
Added Disabled prop for Autocomplete component , also updated story vue.


netchampfaris:
@suhail135 Autocomplete is deprecated, please use Combobox.

PR #387, opened 2025-08-25
feat: add switch to dropdown

safwansamsudeen:
Makes switch a first class component of Dropdown.

This allows users to, for example, keep the dropdown open after clicking:

https://github.com/user-attachments/assets/75b420ae-2ad4-45ef-a0cd-e7abd6611504

### DIFF
--- a/src/components/Dropdown/Dropdown.vue
+++ b/src/components/Dropdown/Dropdown.vue
@@ -30,150 +30,153 @@
               {{ group.group }}
             </DropdownMenuLabel>
 
-            <DropdownMenuItem
-              v-for="item in group.items"
-              :key="item.label"
-              as-child
-              @select="item.onClick"
-            >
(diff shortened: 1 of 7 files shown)


src/components/Dropdown/Dropdown.vue
```diff
@@ -30,147 +30,168 @@
...
-                :active="false"
+            <template v-for="item in group.items" :key="item.label">
+              <Switch
+                v-if="item.switch"
+                class="!px-2 !py-0 h-7"
+                :label="item.label"
+                label-classes="font-normal cursor-pointer"
+                :icon="item.icon"
+                @change="item.onClick?.($event)"
               />
```
netchampfaris, line 41: @safwansamsudeen 
fighting styles too much is a not a good idea, let's just create a switch component without the label and import it here. merging for now, let's clean it up later.

PR #388, opened 2025-08-26
feat: redone select component

safwansamsudeen:
1. Moves over to reka-ui, instead of using HTML's select compontent
2. Fully aligned with `Combobox` - this acts as combobox without the input.
3. Moves common code in `Combobox` and `Select` to a separate file.

Props: `options`, `multiple`, `placeholder`, `disabled`, and `getLabel` - a function to get the label given value(s).

Example:
image

### DIFF
--- a/src/components/Combobox/Combobox.story.vue
+++ b/src/components/Combobox/Combobox.story.vue
@@ -204,77 +204,8 @@ const state = reactive({
       </div>
     </Variant>
 
-    <Variant title="Multiple Selection - Simple">
-      <div class="p-4">
-        <label class="block text-sm font-medium mb-2"
-          >Multiple Simple Options</label
-        >
-        <Combobox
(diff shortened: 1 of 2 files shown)


src/components/Select/Select.vue
```diff
@@ -1,144 +1,169 @@
...
-  ]
-})
+      <SelectPortal class="w-full">
+        <SelectContent
+          :hideWhenDetached="true"
+          :align="'start'"
+          position="popper"
+          class="z-10 min-w-[--reka-select-trigger-width] mt-1 bg-surface-modal overflow-hidden rounded-lg shadow-2xl"
+        >
+          <SelectScrollUpButton
```
safwansamsudeen, line 114: This behavior feels unpolished, as the scroll buttons are absolute. However, when they were in a separate `div`, it looked bad in groups.

Issue #277, opened 2025-03-19
Filters are not keyboard friendly


ankush:
- filters are not keyboard friendly in CRM/Helpdesk (and other that use same components)
- Type something -> use mouse again -> select operator -> use mouse again -> type value is annoying af

Ideally:

- When I click on filter -> focus on first relevant field that I'll be editing
- Tab should jump to next field (operator)
- Typing should select the operator without mouse usage
- Tab again should jump to next field (value)
- enter should apply filter.

ankush:
Also `%%` shouldn't be shown in UI

PR #433, opened 2025-11-03
feat: frappe filter

siduck:
Related to : https://github.com/frappe/frappe-ui/issues/277
- Use combobox instead of Autocomplete
- Supports multiple rows of same field just like desk's filter 
- Filter param output will be an array instead of object

## Screenshot 
 

image

## Usage

```vue
<script setup lang="ts">
import { ref } from "vue";
import { Filter } from "frappe-ui/frappe";

const foo = ref([]);
</script>

<template>
  <main class="grid gap-5 m-20 p-3 rounded">
    {{ foo }}
    <Filter v-model="foo" doctype="ToDo" />
  </main>
</template>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frappe-ui",
-  "version": "0.1.212",
+  "version": "0.1.213",
   "description": "A set of components and utilities for rapid UI development",
   "main": "./src/index.ts",
   "type": "module",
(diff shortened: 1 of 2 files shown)


the-bokya:
I want multiple filters for the same field!

shariquerik (approved):

PR #456, opened 2025-11-16
fix: export sidebaritem

rehanrehman389:


rehanrehman389:
@siduck @shariquerik

siduck:
@rehanrehman389 sidebaritem is directly used in sidebar.vue,  do you import sidebarItem component separately?

Issue #137, opened 2024-04-01
Bug: Select with prefix


shariquerik:
<img width="253" alt="image" src="https://github.com/frappe/frappe-ui/assets/30859809/5f084386-1212-402f-b29e-ab24fd6a320e">

```
<FormControl
  type="select"
  :options="options"
  v-model="value"
>
  <template #prefix>
    <IndicatorIcon />
  </template>
</FormControl>
```

PR #466, opened 2025-11-20
Refactor(Select): move to Reka-ui

siduck:
Closes: #137

image


surajshetty3416 (changes requested):
Screenshot 2025-11-25 at 12 30 45 PM

@siduck you'll have to check dark mode styles for this once... also share SS with examples of custom prefix?

PR #472, opened 2025-11-25
chore: deprecated warning message for Autocomplete

### LINKED ISSUE #434: Deprecate Autocomplete
- Add deprecation message
- Guideline for migrating from Autocomplete to Combobox

siduck:
Closes: #434

image

https://github.com/frappe/frappe-ui/wiki/Guide-to-Migrate-Autocomplete-component-to-Combobox-&-MultiSelect

### DIFF
--- a/src/components/Autocomplete/Autocomplete.vue
+++ b/src/components/Autocomplete/Autocomplete.vue
@@ -210,7 +210,7 @@ import {
   ComboboxOption,
   ComboboxOptions,
 } from '@headlessui/vue'
-import { computed, nextTick, ref, watch } from 'vue'
+import { computed, nextTick, ref, watch, onMounted } from 'vue'
 import { Popover } from '../Popover'
 import { Button } from '../Button'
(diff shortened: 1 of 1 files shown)


surajshetty3416 (changes requested):
@siduck, you should ideally replace usages of `Autocomplete` [example](https://github.com/frappe/frappe-ui/blob/2913bab3d0793e964491711c3a167e21998e03de/src/components/ListFilter/SearchComplete.vue#L2) inside frappe-ui with `Combobox`. Otherwise, people using frappe-ui will keep getting these warnings, and they can’t do anything about it since it’s internal.

netchampfaris:
Closing this. `main` already emits a one-time deprecation warning for Autocomplete via `warnDeprecated` (see `Autocomplete.vue`), so the console.warn here is redundant. The migration guidance it pointed at now lives in a dedicated docs page: #743 (Autocomplete → Combobox / MultiSelect, with before/after tables).

PR #488, opened 2025-12-01
fix(createResource): onError() called thrice

### LINKED ISSUE #67: bug: `onError` in resource is getting called twice
## Sample code
```javascript
createResource({
  url: "https://example.com",
  auto: true,
  validate() {
    return "Sample error";
  },
  onError(error) {
    alert(error);
  },
});
```

## Expected
One alert with 'Sample error'

## Result
Gets two alerts with 'Sample error'

## Possible cause
[handleError](https://github.com/frappe/frappe-ui/blob/b11c531630d335005c1e76deb414e1d8ac264cc7/src/resources/resources.js#L150-L162) is getting called twice from [here](https://github.com/frappe/frappe-ui/blob/b11c531630d335005c1e76deb414e1d8ac264cc7/src/resources/resources.js#L84-L97). The second call might be due to [error thrown](https://github.com/frappe/frappe-ui/blob/b11c531630d335005c1e76deb414e1d8ac264cc7/src/resources/resources.js#L161) from `handleError`

## Possible fix
Remove error thrown from `handleError`?

siduck:
Closes: #67

onError() is already being called inside `resources.js`, no need to call it in  `frappeRequest.js`

https://github.com/user-attachments/assets/6a9073ac-f92c-4064-8901-327d17b93e26

### DIFF
--- a/src/utils/frappeRequest.js
+++ b/src/utils/frappeRequest.js
@@ -100,12 +100,10 @@ export function frappeRequest(options) {
             ? [error._error_message]
             : ['Internal Server Error']
         }
-        options.onError && options.onError(e)
         throw e
       }
     },
     transformError: (error) => {
-      options.onError && options.onError(error)
       throw error
     },
   })


netchampfaris:
@siduck 

This fix might not be correct. frappeRequest is supposed to handle onerror callback if used separately. maybe we add a flag if it is handled already in createResource.

netchampfaris:
Closing as the underlying bug (#67) was already fixed on `main` by 87d983470 (`fix(resources): remove double onError call`) — `onError` now fires exactly once per error through the resource layer.

On the change itself: `resources.js` strips `onError` (`onError: undefined`) before calling the fetcher, so the two lines removed here were already dead code on the `createResource` path. For anyone calling `frappeRequest({ onError })` directly, though, removing both calls would drop the callback entirely (every error funnels through `transformError`, so it'd go from being called to never called). Not needed for #67, so closing — happy to revisit separately if direct `frappeRequest` error handling needs tightening.

PR #494, opened 2025-12-04
feat(createResources): abort method to cancel ongoing fetch requests

### LINKED ISSUE #26: Cancel ongoing fetch requests when component is unmounted

siduck:
Closes: #26

## Usage   

```js
const resp = createResources()

unMounted(() => {
     resp.abort()
})
```

https://github.com/user-attachments/assets/b573971d-d8d6-4f8a-8168-c0bdec50ce7f

To those who want to do this automatically , just make a wrapper over createResources()

```js

const wrapperFetch = (opts) => {
  const resp = createResource(opts)
  onUnmounted(resp.abort)
  return resp
}
```

Docs preview: https://ui.frappe.io/pr-preview/pr-494/

Coverage: 49.34% (+0.02% vs `main`)

### DIFF
--- a/src/resources/resources.js
+++ b/src/resources/resources.js
@@ -31,6 +31,8 @@ export function createResource(options, vm) {
     ? debounce(fetch, options.debounce)
     : fetch
 
+  const controller = new AbortController()
+
   let out = reactive({
     method: options.method,
     url: options.url,
@@ -45,6 +47,7 @@ export function createResource(options, vm) {
     fetch: fetchFunction,
     reload: fetchFunction,
     submit: fetchFunction,
+    abort: () => controller.abort(),
     reset,
     update,
     setData,
@@ -94,6 +97,8 @@ export function createResource(options, vm) {
         return
       }
     }
+
+    options.signal = controller.signal
 
     try {
       out.promise = resourceFetcher({
--- a/src/resources/resources.js
+++ b/src/resources/resources.js
@@ -31,7 +31,7 @@ export function createResource(options, vm) {
     ? debounce(fetch, options.debounce)
     : fetch
 
-  const controller = new AbortController()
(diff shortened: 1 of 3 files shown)


siduck:
> The dev has to write these unmount wrappers manually? 👀
> 
> Can we do a plugin (or hook) or something which automatically aborts? I think it should be possible at least for route changes.

hm so the user can just use this wrapper of  createResource

```js
import { abortableFetch } from "frappe-ui"

const posts = abortableFetch( { url } )
```

netchampfaris:
**Concerns** — the single `AbortController` isn't reusable after an abort.

- `src/resources/resources.js:34` — `controller` is created once per resource. An `AbortController`'s signal is permanently aborted after `abort()`; since `fetch()` always reuses `controller.signal` (`:101`), the first `abort()` kills every subsequent `fetch`/`reload`/`submit` — they reject immediately with `AbortError`. The controller needs to be recreated per fetch (e.g. assign a fresh `controller` at the top of `fetch()`, with `abort()` closing over the current one).
- `src/resources/resources.js:101` — an aborted promise rejects into the `catch` → `handleError`, so a deliberate `abort()` lands in `out.error` as an `AbortError`. Worth short-circuiting `AbortError` so a cancel doesn't surface as a resource error.
- No test for the abort path; given the reuse trap above, a test that aborts then refetches would catch it.

---

**Update** — pushed both fixes to `fetch-abort`:

- `d9e93f3` — recreate the `AbortController` per `fetch()` so the resource stays usable after an `abort()`; skip `AbortError` in the catch so a deliberate cancel doesn't surface as `out.error`.
- `54481dc` — `src/resources/resources.test.ts` covering the abort path and reuse-after-abort. The reuse test fails against the single-controller version, so it guards the regression. Both pass locally.

Issue #515, opened 2025-12-24
Show selected item indicator in the new Select field


ruchamahabal:
Added in: https://github.com/frappe/frappe-ui/pull/466

Native Select input has an indicator for current selection:

<img width="262" height="139" alt="Image" src="https://github.com/user-attachments/assets/ca28c847-b154-4fc9-b21d-9092826e8637" />

The new Select field doesn't. It's a bit disorienting

<img width="274" height="175" alt="Image" src="https://github.com/user-attachments/assets/1471e68a-0a5c-45bf-a469-bfec6d8d5798" />

PR #518, opened 2025-12-29
feat(UX): add selected item indicator in Select component

ruchamahabal:
Closes https://github.com/frappe/frappe-ui/issues/515

**Before**:
image

**After:**
image


siduck:
@ruchamahabal why not just make use of flex container by making the check icon have ml-auto instead of absolute positioning https://github.com/frappe/frappe-ui/pull/519/files

ruchamahabal:
I just followed reka-ui's documentation implementation https://reka-ui.com/docs/components/select#implementation & looked at Combobox's code. Anything is fine

siduck:
If you wrap the icon inside the `<SelectItemIndicator>` then there will a wrapper container  and then you're adding flex & items-center to center it

image

I think its un-necessary and we could pass the check icon directly and it will be centered cuz its parent div has flex already
 https://github.com/frappe/frappe-ui/blob/73fa1116490ce442628d318e2d768376d36da4c2/src/components/Select/Select.vue#L111

siduck (approved):

PR #523, opened 2026-01-01
feat: add hovercard component

LeoDanielA01:
## What does this PR do?

Introduces a new **HoverCard** component to Frappe UI for displaying
rich, contextual information on hover or keyboard focus.

---

## Why is this needed?

Current tooltips in Frappe UI are suitable only for short text.
Many ERP use cases require richer contextual previews such as:

- User / owner information
- Inline help with formatting
- Metadata previews
- Contextual explanations without opening dialogs

HoverCard fills this gap without adding modal overhead.

---

## Features

- Hover and keyboard focus triggers
- Accessible by default (focusable, non-blocking)
- Supports rich content
- Lightweight alternative to dialogs
- Reusable and composable API
- No breaking changes


netchampfaris:
https://ui.frappe.io/docs/components/hovercard

We now have HoverCard

PR #522, opened 2026-01-01
feat: Add context menu component

LeoDanielA01:
## What does this PR do?

Adds a Context Menu component to Frappe UI, and implemented natively using Vue.

## Why is this needed?

Frappe UI currently lacks a reusable context menu primitive.
This component enables right-click actions for tables, lists, trees,
and other ERP workflows.

## Features

- Right-click trigger support
- Keyboard accessibility
- Menu items, separators
- Clean, reusable API

## Checklist

- No build artifacts committed
- Scoped to a single component
- No breaking changes

PR #524, opened 2026-01-02
feat: add skeleton component

LeoDanielA01:
## What does this PR do?

Improves the Skeleton component to better support
flexible presentation requirements commonly needed
in modern UI patterns.

---

## What is the main gain?

This change removes the need for wrapper elements
or inline style workarounds when using Skeletons,
resulting in cleaner markup and more consistent
usage across Frappe UI.

---

## Why is this needed?

Skeletons are widely used as loading placeholders
across different UI surfaces such as user previews,
lists, and contextual components. Without sufficient
flexibility at the component level, consumers are
forced to rely on implementation-specific hacks.

This PR provides a clean, component-level solution
that improves developer experience while keeping
existing usage intact.

Coverage: 58.54% (+0.05% vs `main`)


siduck:
Attach some images of the UI

LeoDanielA01:
@siduck Could you pls review this PR?. i've also opened few other PR's could review those as well. PR Id : #523  , #522

Issue #525, opened 2026-01-04
enhancement: add underline option in the TextEditor Menu


prathameshkurunkar7:
Having underline option in the editor menu can be a good edition to the TextEditor. Also people have requested this on [Gameplan](https://github.com/frappe/gameplan/issues/396).

Having this in the TextEditor component itself makes more sense. A good enhancement to have.

PR #526, opened 2026-01-04
feat: enhance TextEditor add underline option in the texteditor menu

prathameshkurunkar7:
Solution for - #525


netchampfaris:
Underline is already available in the newer Editor molecule (`src/molecules/editor`), the mark ships in the default kit and works via Cmd/Ctrl+U. We've deliberately left it off the default toolbar (not a great default for most apps); apps can expose a button via their own menu config.

This PR targets the older `TextEditor` component, which is being replaced by the Editor molecule, so closing this one.

PR #533, opened 2026-01-08
test: setup component testing

netchampfaris:
To avoid regressions


netchampfaris:
Component testing added in https://github.com/frappe/frappe-ui/pull/567

PR #538, opened 2026-01-09
fix: tables menus, size handler and copy pasting to new tab

DeVaNsHk72:
Screenshot 2026-01-09 at 9 07 53 PM


netchampfaris:
@safwansamsudeen Table is shipped in Editor, closing.

PR #575, opened 2026-02-04
fix(fileupload): error handling in file upload

raizasafeel:
### Summary
- Move 'finish' to fire after server response instead of after upload completes
- nginx returns HTML, we listen for JSON, which failed silently before, now added checks for that, which returns a message and httpStatus

### Change
- When toasted out
Screenshot 2026-03-27 at 11 29 12 PM

### Edit
- Removed toast

### Example usage in code

  ```vue
 <FileUploader @failure="(error) => toast.error(error?.message || 'Upload failed')"/>
```

### Issue
- Fixes frappe/lms#2267

### DIFF
--- a/.github/workflows/component-test.yml
+++ b/.github/workflows/component-test.yml
@@ -1,0 +1,52 @@
+name: Component Tests
+on:
+  push:
+
+permissions:
+  contents: read
+  actions: write
+
+jobs:
+  component-test:
+    runs-on: ubuntu-latest
+    name: Component Tests (Cypress) - ${{ matrix.containers }}
+    strategy:
+      fail-fast: false
+      matrix:
+        containers: [1, 2, 3]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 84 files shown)

### CODE REVIEW

File: src/components/FileUploader/FileUploader.vue [hunk 1]
```diff
@@ -121,14 +122,24 @@ export default {
...
               JSON.parse(error._server_messages)[0],
             ).message
           } else if (error?.exc) {
             errorMessage = JSON.parse(error.exc)[0].split('\n').slice(-2, -1)[0]
           }
           this.error = errorMessage
+
+          // Show toast only for file size limit errors, console log others
+          if (error?.httpStatus === 413) {
+            toast.error(errorMessage)
```
netchampfaris: Should be in app facing code and not core library? Developer might decide to show the error "inline" instead of a toast. The example you show in the PR description is a good example of this, if the dialog button triggers the upload error, it's better to show the error inside the dialog.

PR #583, opened 2026-02-19
fix(TextEditor): Could not find image node with uploadId

krantheman:
When pasting images without typing or making immediate edits, the image `src` remains as base64 instead of being updated to the uploaded file URL.

This is caused because the content watcher calls `setContent()` while images are still uploading, destroying image nodes with `uploadId` references before the upload callback can update them with the final file URL. 

Adding a check to skip `setContent()` calls when images have the `loading` attribute fixes the issue.


safwansamsudeen:
@krantheman as discussed this is likely not the real problem. Can we close?

PR #633, opened 2026-04-24
feat(Combobox): v1 with button-trigger mode, shared popover motion, and curated demos

netchampfaris:
## Summary

- **Combobox v1** per [`08d-combobox-spec.md`](./v1-release/08d-combobox-spec.md) — new per-item `slots` API (`{ prefix, label, suffix, item }`), intrinsic-width trigger, press feedback, chevron rotation. Deprecations with back-compat aliases: `render`, `slotName`, `input` event, `placement`.
- **Button trigger mode** (new `trigger: 'input' | 'button'` prop): renders a built-in `Button`, moves the search input into the popover header and auto-focuses it on open. Label + prefix auto-wire from the selected option (reuses `#item-prefix`, falls back to `selectedOption.icon`, then `#prefix` as placeholder). `#trigger` slot remains for fully custom triggers.
- **Shared `usePopoverMotion`** composable — Combobox, Select, and Dropdown classify opens as pointer-driven or keyboard-driven via pointer-recency (`@pointerdown` timestamp within 300ms of the open transition). Keyboard opens skip the scale entrance and run an 80ms opacity-only fade to mask reka's 1-frame position-settle; pointer opens play the full animation.
- **Stories**: 6 curated demos — Simple (repo picker), EmojiPicker (button mode), Grouped (spaces by team), CustomValue, StatusPicker, MemberPicker.
- **Docs**: `<ComponentPreview layout="stacked">` shows preview + code together; the markdown plugin now forwards arbitrary props instead of whitelisting.
- **Tests**: 36 passing (up from 7).
- **Spec**: main RFC split into per-component sub-specs (`08a`–`08e`); new shared rule 10 "Popover motion conventions".

## Test plan

- [x] Combobox.cy.ts — 36/36 pass
- [x] Select.cy.ts — 15/15 pass (adopted shared composable)
- [x] Dropdown.cy.ts — 6/6 pass (adopted shared composable)
- [x] ItemList.cy.ts / ItemListRow.cy.ts — 10/10 pass (regression check)
- [x] `yarn docs:build` succeeds
- [ ] Manual verification on dev server: keyboard-tab into combobox is instant (no animation); click to open animates; button-trigger (EmojiPicker story) opens with search focused in popover; `trigger="button"` reuses `#item-prefix` for the selected-state button prefix and `#prefix` as the placeholder icon.

## Summary by CodeRabbit

* **New Features**
  * Added customizable trigger modes and enhanced slot system for Combobox
  * Introduced new Combobox variants: Emoji Picker, Status Picker, and Member Picker
  * Added layout flexibility for demo components (tabs vs. stacked)
  * Improved motion and animation handling across picker components

* **Documentation**
  * Expanded component API documentation with detailed specs
  * Added comprehensive examples and migration guides

* **Tests**
  * Added extensive Combobox test coverage for interactions and accessibility


netchampfaris:
Addressed CodeRabbit review in `87c6394e`:

**In-scope fixes applied:**

| # | File | Severity | Fix |
| --- | --- | --- | --- |
| 8 | `Combobox.vue:414` | 🔴 Critical | `ComboboxTrigger` now always uses `as-child` in button mode. Previously the built-in `<Button>` nested inside ComboboxTrigger's default `<button>`, producing invalid HTML. Added a regression test asserting `[data-slot="trigger"] button` doesn't exist. |
| 6 | `Combobox.vue:70` | 🟠 Major | Dropped the `passive: (...) as false` type cast. `useVModel` now uses `passive: true` unconditionally so the return type stays a plain `Ref` regardless of controlled/uncontrolled usage. |
| 9 | `ComboboxResults.vue:250` | 🟠 Major | Precedence comment rewritten to match the code: per-item `#item-<slot>` dispatch takes precedence over the generic `#item` fallback, as the `shouldUseDynamicItemSlot` guard enforces. |
| 7 | `Combobox.vue:307` | 🟡 Minor | `reset()` now emits `update:query` with `''` so `@update:query` consumers see the clear transition alongside `update:modelValue` and `update:selectedOption`. Test updated to assert it. |
| 3 | `docs/meta/Combobox.md:133` | 🟡 Minor | `placement` prop now has a proper JSDoc summary in `types.ts`, so the auto-generated table shows a real description instead of an empty one. Still tagged `@deprecated`. |

**Not addressed (out of scope for this PR):**

| # | File | Note |
| --- | --- | --- |
| 1 | `docs/.vitepress/plugins/componentTransformer.ts:84` | Attribute escaping. Current call sites pass plain strings; no value contains `"` or HTML entities today. Will fix if/when a real callsite needs it. |
| 2 | `docs/meta/Calendar.md:50` | Slot-type truncation in `propsgen.ts`. Affects Calendar and Popover, unrelated to Combobox. Worth a follow-up PR to lift the 100-char cap. |
| 4 | `docs/meta/ListFilter.md:22` | ListFilter uses runtime props without TS generics. Source component refactor, separate PR. |
| 5 | `docs/meta/Toast.md:68` | `action` emit missing description. Lives in Toast source, unrelated. |

The three deferred comments (2, 4, 5) all touch components outside this PR's scope — they surfaced because I regenerated `docs/meta/*.md` as a side effect and the existing source files have the same issues on `main`. Comment 1 is a defensive hardening that doesn't fix any known-broken call site.

Tests: 37/37 pass (+1 nested-button guard, +1 `update:query` on reset).
Docs: `yarn docs:build` clean.

PR #680, opened 2026-05-16
feat: use `Button` component in `SidebarLink`

nextchamp-saqib:
## Coverage

| Lines      | 41.53% (1658/3992) | ±0.00%      |
| Statements | 36.55% (1734/4743) | ±0.00%      |
| Functions  | 37.52% (334/890) | ±0.00%      |
| Branches   | 26.08% (474/1817) | ±0.00%      |


nextchamp-saqib:
`SidebarLink` uses `Button` but it doesn't use the button's styles or slots, and expand-collapse transition is not implemented in button, so it looks weird. 

Thought of making these changes in `Button` itself, but after chat with @netchampfaris  decided to keep `SidebarLink` separate. 

So will be refactoring `SidebarLink` to not use `Button` and then probably merge

PR #711, opened 2026-05-19
fix(PickerShell): pass MouseEvent in click emit

ruchamahabal:
Clicking on a DatePicker breaks in Studio. Studio needs to show the editor outline on the clicked component, so it has a click event handler. When PickerShell emits 'click', it doesn't pass the MouseEvent. Since DatePicker doesn't declare click in its emits, `@click="handleClick"` from StudioComponent falls through to PickerShell as a component event listener. When PickerShell's onClick() calls emit('click') without args, handleClick receives undefined instead of a MouseEvent, causing it to break on click

**Before**:

image

**After**:

image

## Coverage

| Lines      | 44.04% (1889/4289) | ±0.00%      |
| Statements | 38.96% (2001/5135) | ±0.00%      |
| Functions  | 39.83% (378/949) | ±0.00%      |
| Branches   | 29.42% (595/2022) | ±0.00%      |

### DIFF
--- a/src/components/shared/picker/PickerShell.vue
+++ b/src/components/shared/picker/PickerShell.vue
@@ -123,7 +123,7 @@ const props = withDefaults(defineProps<Props>(), {
 
 const emit = defineEmits<{
   (e: 'focus'): void
-  (e: 'click'): void
+  (e: 'click', event: MouseEvent): void
   (e: 'blur'): void
   (e: 'enter'): void
   (e: 'open'): void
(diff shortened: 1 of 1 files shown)


src/components/shared/picker/PickerShell.vue
```diff
@@ -123,7 +123,7 @@ const props = withDefaults(defineProps<Props>(), {
 
 const emit = defineEmits<{
   (e: 'focus'): void
-  (e: 'click'): void
+  (e: 'click', event: MouseEvent): void
```
ruchamahabal, line 126: On second thought, do we even need to define this as an emit @netchampfaris? None of the components using PickerShell are listening to `@click`.
If we don't define this emit, event handling will work as a native listener

netchampfaris: @ruchamahabal I think it's needed. Also, I dont like this implicit attr forwarding, need to decide whether we should do this explicitly everywhere. We need broad guidelines for attr forwarding.

PR #720, opened 2026-05-21
feat: export tailwindconfig tokens

ruchamahabal:
Move inline tokens to tailwind/tokens.js and export them from package.json to allow other projects to import and use them. Need this for Studio [ref](https://github.com/frappe/studio/pull/175/changes/0dd99df4d590f326892b64487e41b043c924c0cf)

I was using resolveConfig to do this earlier, but now it breaks because the preset in tailwindconfig imports `lucideIconsPlugin.js`, which uses createRequire - a Node-only API that doesn't work in the browser. Also, it was not ideal to do so. Tailwind suggests writing tokens in a js file and using/exposing them for external use.

If there is a better way to do this, please let me know

## Coverage

| Lines      | 48.92% (5047/10315) | ±0.00%      |
| Statements | 46.18% (5160/11173) | ±0.00%      |
| Functions  | 47.13% (1470/3119) | ±0.00%      |
| Branches   | 40.14% (2359/5876) | ±0.00%      |

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,9 @@
       "import": "./tailwind/index.js",
       "default": "./tailwind/index.js"
     },
+    "./tailwind/tokens.js": {
+      "import": "./tailwind/tokens.js"
+    },
     "./vite": {
       "import": "./vite/index.js"
     },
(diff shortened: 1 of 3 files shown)


netchampfaris:
@ruchamahabal It's fine for now. Will have to find a better way when we migrate to Tailwind 4.

PR #759, opened 2026-06-05
fix(docs): resolve color token values so swatches render in docs pages

### LINKED ISSUE #760: Color swatches render empty on Design System docs pages
On the docs site, the color swatches on the Design System pages render as empty/transparent boxes, only the section headings and labels show up, no actual colors.

  **Steps to reproduce:**

  1. `yarn docs:dev`
  2. Open *Design System → Background Color* (also *Border Color* and *Text Design → Text Color* tab)

  **Actual:** swatches are invisible (transparent), only labels render

  **Expected:** each swatch shows its token color (e.g. `bg-surface-red-4`)

  **Version:** `main` @ v1.0.0-beta.3

Image

  **What I found:** the token values passed by `docs/content/docs/design-system/[token].paths.ts` contain Tailwind's internal `<alpha-value>` placeholder (e.g. `color-mix(in srgb, var(--surface-gray-2, #F3F3F3) calc(<alpha-value> * 100%), transparent)`), which is invalid as an inline style — so the swatch `div`s get no background.

  I have a small fix for this in #759.

maarkN:
Fixes #760

## Problem

  On the docs Design System pages (`yarn docs:dev` → Background Color, Border Color,
  Text Design › Text Color), all color swatches render transparent/empty — the colors
  don't show up.

  ## Root cause

  The semantic color tokens generated in `tailwind/colorPalette.js` resolve to Tailwind
  strings containing the internal `<alpha-value>` placeholder, e.g.:

  ```
  color-mix(in srgb, var(--surface-gray-2, #F3F3F3) calc(<alpha-value> * 100%), transparent)
  ```

  That placeholder is only substituted by Tailwind when generating utility classes.
  `docs/content/docs/design-system/[token].paths.ts` was passing this raw value to the
  token components, which use it as an inline style (`backgroundColor` / `color` /
  `borderColor`) — invalid CSS, so the swatches rendered transparent.

  ## Fix

  Added a small resolver in `[token].paths.ts` that extracts the theme-aware
  `var(--token, fallback)` portion of the value and applied it in the three color
  getters (`getBgColors`, `getTextColors`, `getBorderColors`). Since the CSS variables
  are defined on `:root` / `[data-theme="dark"]` by the theme plugin, the swatches also
  follow dark mode correctly.

  ## Before / After

06-before-after-background-color
 
 Dark mode (swatches follow the theme):
05-after-background-color-dark

 Also fixes Border Color and Text Design › Text Color, which used the same data:
 
03-after-border-color
04-after-text-design

### DIFF
--- a/docs/content/docs/design-system/[token].paths.ts
+++ b/docs/content/docs/design-system/[token].paths.ts
@@ -3,6 +3,15 @@ import config from '../../../../tailwind.config'
 
 const designTokens = resolveConfig(config).theme
 
+/**
+ * Semantic color tokens resolve to Tailwind strings with the internal `<alpha-value>` placeholder,
+ * e.g. `color-mix(in srgb, var(--surface-gray-2, #F3F3F3) calc(<alpha-value> * 100%), transparent)`,
+ * which is invalid as an inline style. Extract the theme-aware `var(--token, fallback)`
+ * part so swatches render (and follow dark mode).
+ */
+const resolveColorValue = (value: string): string =>
+  value.match(/var\([^)]*\)/)?.[0] ?? value.replace(/<alpha-value>/g, '1')
+
 const getBgColors = () => {
   let colors: { name: string; value?: string }[] = []
   const list = designTokens.backgroundColor.surface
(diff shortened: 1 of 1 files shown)


netchampfaris:
This is now fixed

PR #763, opened 2026-06-08
fix: export `internals` endpoint 

RitvikSardana:
Since we are now shipping backend-related components in the framework. 
We would need to access internal composables / utils / types.
Hence, we need to expose those from frappe-ui 

example 
https://github.com/frappe/frappe/pull/39810/changes#diff-7d8a4747baa68c8ac0d76ddca9d0d9f6a512c6396c080f99d727c523d340ae7bR133-R134

Coverage: 56.48% (±0.00% vs `main`)

image


RitvikSardana:
We can merge this right @netchampfaris ?

shariquerik:
@RitvikSardana The docs for these components should be on a different path -> `/docs/internals/components`

shariquerik:
@RitvikSardana, I think Sidebar should only have a link to internals, which will rerender the sidebar with internals components/utilities

shariquerik:
@RitvikSardana, what about phone controls docs?

RitvikSardana:
That we still need to figure out, as those are not exposed by frappe-ui, it is exposed by framework @shariquerik

PR #768, opened 2026-06-09
fix: use frappe.client.save in save() to fix Single DocType TypeError

### LINKED ISSUE #767: fix: save() fails with TypeError: dict can not be used as parameter for Single DocTypes
**Versions**
- frappe-ui: 0.1.261
- frappe: 15.110.0
- crm: 1.73.0

**Describe the bug**
Calling `document.save.submit()` on a Single DocType (e.g. CRM Twilio Settings) 
fails with:

TypeError: dict can not be used as parameter

**Root Cause**
In `documentResource.js`, the `save` resource uses `frappe.client.set_value` 
and passes the entire doc object as `fieldname`:

```js
return {
  doctype: out.doctype,
  name: out.name,
  fieldname: values, // entire doc object — crashes for Single DocTypes
}
```

frappe.client.set_value cannot handle a dict as fieldname parameter.

**Fix**
Use `frappe.client.save` instead, which accepts the full doc correctly.
PR attached.

prateekjoshi44:
Fixes #767 
## Problem
Calling `document.save.submit()` on a Single DocType (e.g. CRM Twilio Settings) fails with:
TypeError: dict can not be used as parameter

## Root Cause
In `documentResource.js`, the `save` resource uses `frappe.client.set_value` 
and passes the entire doc object as `fieldname`:

```js
save: createResource({
  ...setValueOptions, // inherits url: frappe.client.set_value
  makeParams() {
    let values = JSON.parse(JSON.stringify(out.doc))
    return {
      doctype: out.doctype,
      name: out.name,
      fieldname: values, // ← entire doc object passed as fieldname
    }
  },
})
```

`frappe.client.set_value` cannot serialize a dict as the `fieldname` 
parameter — crashes specifically for Single DocTypes.

## Fix
Replaced `save` to use `frappe.client.save` which accepts the full doc correctly.

## Tested On
- frappe-ui: 0.1.261
- frappe: 15.110.0
- crm: 1.73.0
- Affected: CRM Twilio Settings, Exotel Settings and any Single DocType save

### DIFF
--- a/src/resources/documentResource.js
+++ b/src/resources/documentResource.js
@@ -1,9 +1,9 @@
 import { reactive, watch, effectScope } from 'vue'
 import { getCacheKey, createResource } from './resources'
 import {
-  updateRowInListResource,
-  deleteRowInListResource,
-  revertRowInListResource,
+    updateRowInListResource,
+    deleteRowInListResource,
+    revertRowInListResource,
 } from './listResource'
 import { getLocal, saveLocal, deleteLocal } from './local'
 import { onDocUpdate } from './realtime'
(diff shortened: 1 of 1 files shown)


netchampfaris:
Thanks for catching this and for the fix, @prateekjoshi44!

This turned out to already be resolved on `main`. The underlying problem — a Single DocType's `save()` failing because read-only standard fields end up in the `set_value` payload — was fixed by the recent "send only changed fields" change: `save` now sends `getChangedFields()`, which diffs against the loaded doc and strips `doctype`/`name`/meta fields.

I verified against a live Single DocType (with rollback):

- Current `main` — `set_value(dt, dt, {changed_field: value})` on a Single → **works**
- Old path — whole doc as `fieldname` on a Single → **fails** with `Cannot edit standard fields`
- Normal (non-Single) doctype, current shape → works (no regression)

This PR branched from before that change, so its diff would revert `getChangedFields()` (and the skip-if-unchanged optimization) back to sending the whole doc via `frappe.client.save`, and it now conflicts with `main`.

Closing as already fixed. If you still hit this on the latest `main` (e.g. with CRM Twilio Settings), please reopen with the steps and we'll dig in. Appreciate the report!

Issue #756, opened 2026-06-04
toast issues: v1 beta 


RitvikSardana:
1. If there are two toasts and one of the toasts is multiline and one is not, the collapsed state looks ugly

<img width="424" height="184" alt="Image" src="https://github.com/user-attachments/assets/9a279d49-7746-411b-a017-309c3f0767c7" />

<img width="403" height="109" alt="Image" src="https://github.com/user-attachments/assets/d79af39c-40bd-46d4-8dff-08412c86cdd0" />

2. HTML does not render properly inside the toast; previously, we had "safe" tags.

PR #769, opened 2026-06-09
fix: render safeHTML in toasts

RitvikSardana:
## 1. Render safe HTML inside the toast

Backend-supplied HTML was showing up as escaped text instead of rendering. We now sanitize the message with **DOMPurify** (already in `package.json`) and render it — only the allow-listed tags survive, everything else (scripts, event handlers, unknown tags) is stripped.

**How it works:**

```js
const ALLOWED_TAGS = ['a', 'em', 'strong', 'i', 'b', 'u']

1. toast.success('<b>Saved</b>') — message comes in as a string
2. DOMPurify.sanitize(message, { ALLOWED_TAGS }) — strip unsafe tags
3. () => h('span', { innerHTML: cleanHTML }) — wrap as a render fn (≈ v-html)
4. Pass the render fn to vue-sonner as the title
5. Sonner's Toast.vue renders it via <component :is="toast.title" /> 
```

## 2. Fix the collapsed state with mixed toast heights

When two toasts were stacked in the collapsed state — one multiline, one single-line — the taller toast's content spilled out the top of the card and looked broken.

vue-sonner clamps every non-front toast to the front toast's height and hides their content, but only on [data-styled] toasts — and our unstyled config strips that attribute. We re-add the rule so collapsed back toasts fade their content out, matching sonner's intended stacked look.

Image

Image

`closes`: https://github.com/frappe/frappe-ui/issues/756

Coverage: 56.93% (+0.04% vs `main`)

### DIFF
--- a/src/components/Toast/Toast.sanitize.test.ts
+++ b/src/components/Toast/Toast.sanitize.test.ts
@@ -1,0 +1,50 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Sanitization is verified against the REAL DOMPurify (needs a DOM, hence the
+ * jsdom environment) — the allow-list wiring is covered separately in
+ * Toast.test.ts with a stubbed DOMPurify.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import type { VNode } from 'vue'
+
+const sonnerSpy = Object.assign(vi.fn(), {
+  success: vi.fn(),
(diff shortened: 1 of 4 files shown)


netchampfaris:
/barista review does the sanitize test need to be in a separate file?

RitvikSardana:
@netchampfaris 
Different files seem fine, right?
If yes, we can merge this

PR #770, opened 2026-06-09
fix: add seaction-header slot

RitvikSardana:
Add `section-header` slot in `SidebarSection` component.

By default the heading is rendered like this (chevron icon on the right of the heading):
image

We can customize it accordingly, like
image

Coverage: 56.89% (±0.00% vs `main`)

### DIFF
--- a/src/components/Sidebar/Sidebar.vue
+++ b/src/components/Sidebar/Sidebar.vue
@@ -25,6 +25,9 @@
       :items="section.items"
       :collapsible="section.collapsible"
     >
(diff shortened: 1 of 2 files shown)


netchampfaris:
@RitvikSardana Molecule components shouldnt be customizable via slots but via composition of smaller components. Just like Editor molecule. This needs to be rewritten.

RitvikSardana:
You mean sidebar component needs to be re-written? @netchampfaris

RitvikSardana:
Closing since we have a new sidebar component now

PR #776, opened 2026-06-11
fix: remove default shadow from Card component

### LINKED ISSUE #735: Card should not have shadow by default
Looks inconsistent

Image

mvanhorn:
## Summary

`Card` no longer hardcodes the `shadow` class on its root element. The card keeps its border, rounding, and background; the drop shadow that made it look inconsistent next to flat-styled cards is gone, as requested in #735 ("But it does ship with one, just remove the shadow?").

No prop was added - the directive in the issue is removal. An opt-in `shadow` prop can follow later if review wants one.

Fixes #735

Coverage: 56.48% (±0.00% vs `main`)

### DIFF
--- a/src/components/Card.vue
+++ b/src/components/Card.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex flex-col rounded-lg border bg-white px-6 py-5 shadow">
+  <div class="flex flex-col rounded-lg border bg-white px-6 py-5">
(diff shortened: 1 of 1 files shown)


mvanhorn:
Really appreciate the quick merge, @netchampfaris. Card keeps its border and rounding without the hardcoded shadow now.

PR #787, opened 2026-06-16
feat(FloatingWindow): add dockable floating window component

aerodeval:
## Demo

https://github.com/user-attachments/assets/b44c9b32-1e84-441b-a122-b78b7fb749c9

## What

Adds a **`FloatingWindow`** component: a panel that docks inline, pops out into a draggable, resizable floating window, or minimizes to a bottom-right tray. It is the interaction pattern of a chat panel or mail composer.

While detached it teleports to `<body>`, so it escapes any host layout's overflow and stacking contexts (Teleport moves the node without remounting, so an editor or form inside keeps its focus and state). It can persist its geometry across sessions via an optional `storage-key`.

Only one window is detached at a time: popping one out docks whichever window was already floating, the way a single chat or mail composer behaves.

Also exports the headless **`useFloatingWindow`** composable, the drag, resize, and window-state engine on its own, for building a custom shell.

## Why

Apps repeatedly hand-roll "pop-out composer" windows (reply boxes, chat panels). This packages the fiddly parts (viewport clamping, the docked / floating / minimized state machine, geometry persistence, and the popover z-index fix) behind a small, reusable surface.

## API

- Uncontrolled by default; the built-in title bar drives the state. Bind `v-model:mode` to control it from the host (`docked | floating | minimized`).
- Props: `title`, `storageKey`, `initialWidth`, `initialHeight`, `minWidth`, `minHeight`.
- Slots: `default` (the scrollable body), `footer` (a pinned composer or action bar), `header` (replaces the title bar entirely and becomes the drag handle), `actions` (extra controls beside the built-in window buttons).

## Notes

- Reuses existing dependencies: VueUse (`useDraggable`, `useStorage`, `useWindowSize`) and the in-repo `Button`. No new deps.
- A `body.has-floating-window` class lifts the reka-ui popover layer (`[data-reka-popper-content-wrapper]`) above the window, so dropdowns and autocompletes opened inside a floating window stay clickable.
- The component owns the "body scrolls, footer pins" layout, so consumers do not hand-roll it.
- Ships with six end-to-end interactive stories (chat, custom header, actions, persisted EmailComposer-style draft, controlled, headless) and an auto-generated API table on the docs page.

## Relocation & breaking change

`FloatingWindow` now ships under the unstable **`frappe-ui/experimental`** subpath rather than the public API, so its surface can keep settling without a deprecation promise. It stays importable, but is no longer exported from the package root or shown in the component docs.

```ts
// before
import { FloatingWindow, useFloatingWindow } from 'frappe-ui'
// after
import { FloatingWindow, useFloatingWindow } from 'frappe-ui/experimental'
```

> [!WARNING]
> **The `frappe-ui/internals` subpath has been renamed to `frappe-ui/experimental`.** This is a breaking change: `frappe-ui/internals` no longer resolves. If you import from it (`useInputLabeling`, `InputLabel`, `InputDescription`, `InputError`, `LabelingWrapper`, `inputFontSizeClasses`, and now `FloatingWindow`), update the specifier to `frappe-ui/experimental`. No other change is required; the exports are identical.

Coverage: 57.98% (±0.00% vs `main`)


NagariaHussain:
Demo video? 👀

netchampfaris:
@aerodeval we should rename `internals` to `experimental`, internals doesnt feel right and doesnt capture the intent. We dont have many APIs here yet to better to rename now.

PR #800, opened 2026-06-26
fix(useDoc): don't self-evict the doc ref on first uncached load

netchampfaris:
## Problem

The first uncached load of any document via `useDoc` crashes and renders blank:

```
Failed to set doc in IDB: TypeError: Cannot read properties of undefined (reading 'value')
  at useDoc.ts   → let value = storeRef.value
  at docStore.ts → docRef.value = doc   (setDoc)
```

## Root cause — two defects, both from `getDoc` returning the in-memory ref synchronously

`getDoc` returns the ref synchronously and its caller (useDoc's `doc` computed) dereferences it immediately. Two separate paths violate that contract:

**1. `setDoc` ordering → first-uncached-load crash + IDB churn**

`setDoc` set `lastFetched` *after* `docRef.value = doc`. Assigning the ref synchronously re-runs the `doc` computed, which re-enters `getDoc`; the entry still looked stale (`lastFetched` unset), so `getDoc` kicked off a reload that — via `cleanup()` — deleted the very ref it was about to return. `getDoc` then returned `undefined` and the computed crashed on `.value`. `loadDoc(firstLoad=true)` never sets `lastFetched` when IDB is empty, so this fired on **every first uncached load**.

**2. `loadDoc` stale path → stale-access crash (independent of #1)**

The stale-reload branch called `cleanup()`, whose first synchronous line is `this.docs.delete(key)`. So any doc accessed after it goes stale (a long-lived `useDoc` re-evaluated past the cache timeout) makes `getDoc`'s `return this.docs.get(key)!` yield `undefined` and crash — the first-uncached-load was just the most common trigger.

## Fix

1. `setDoc` marks `lastFetched` **before** assigning the ref, so the synchronous re-run sees a fresh entry and never reloads — no crash, and the just-written IDB copy isn't evicted by a needless reload.
2. The `loadDoc` stale path evicts the IDB copy and clears the timestamp but **keeps the in-memory ref**. `cleanup()` is untouched and still used by `invalidateDoc`/`removeDoc`, which intend full removal.

## Tests

Adds `src/data-fetching/docStore.test.ts` (none existed):
- **first uncached load** — a synchronously-tracked reader (mirroring useDoc's computed via `watchSyncEffect`) through `setDoc`; asserts no crash and that the IDB copy survives.
- **stale access** — forces the entry past the cache timeout; asserts `getDoc` never returns `undefined`.

Both fail on the prior code and pass with this change. Full `src/data-fetching/` suite (29 tests) passes.

Reproduced and verified in a downstream app (Gameplan): every first discussion open rendered blank with the error above; with this change the doc loads on first visit, the console is clean, and no unhandled rejection fires.

https://claude.ai/code/session_01CG9r6WdYSLUhcTKbNXcoki

Coverage: 57.93% (±0.00% vs `main`)

### DIFF
--- a/src/data-fetching/docStore.test.ts
+++ b/src/data-fetching/docStore.test.ts
@@ -1,0 +1,73 @@
+/**
+ * @vitest-environment node
+ */
+import { computed, watchSyncEffect } from 'vue'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { docStore } from './docStore'
+import { idbStore } from './idbStore'
+
+const DOCTYPE = 'User'
+const identity = <T>(d: T): T => d
+
+function idbKey(name: string) {
+  return `doc:${DOCTYPE}/${name}`
+}
+
+/**
+ * Mirror useDoc's `doc` computed: it calls getDoc and dereferences the returned
+ * ref *without* optional chaining, so a getDoc that returns undefined throws.
+ * A synchronous watcher reproduces the in-browser render effect that re-runs
+ * this computed the instant setDoc assigns the ref — the trigger for the crash.
+ */
+function subscribeLikeUseDoc(name: string) {
+  const doc = computed(() => docStore.getDoc(DOCTYPE, name, identity).value)
+  const seen: unknown[] = []
+  const stop = watchSyncEffect(() => {
+    seen.push(doc.value)
+  })
+  return { doc, seen, stop }
+}
+
+describe('docStore', () => {
+  beforeEach(async () => {
+    await docStore.clearAll()
+  })
+  afterEach(async () => {
+    await docStore.clearAll()
+  })
+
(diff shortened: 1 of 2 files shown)


netchampfaris:
Updated from review feedback:

- **Killed the IDB churn at its source.** The original eviction-split stopped the crash but left a spurious stale-reload during `setDoc` that wiped the just-written IDB copy. Reordering `setDoc` to mark `lastFetched` before assigning the ref removes that reload entirely (the synchronous re-run now sees a fresh entry), so the cache persists on first load.
- **Kept the `loadDoc` ref-preservation too** — it fixes a *second*, independent crash the reorder doesn't reach: any doc accessed after it legitimately goes stale would still hit `cleanup()`'s synchronous `docs.delete` and make `getDoc` return `undefined`.
- **Added `docStore.test.ts`** (there was no docStore test) covering both the first-uncached-load and stale-access paths. Both fail on the prior code and pass now.

PR #803, opened 2026-06-26
fix(documentResource): send only changed fields on save

### LINKED ISSUE #802: documentResource save sends standard fields and fails with Cannot edit standard fields
## Summary

`documentResource` `save` serializes the whole `out.doc` and strips only `doctype` and `name` so the `frappe.client.set_value` payload still contains `owner` `creation` `modified` `modified_by` `docstatus` `idx`. The request fails with `Cannot edit standard fields` and the edit is never persisted.

## Location

`src/resources/documentResource.js` `save.makeParams`

```js
makeParams() {
  let values = JSON.parse(JSON.stringify(out.doc))
  delete values.doctype
  delete values.name
  return { doctype: out.doctype, name: out.name, fieldname: values }
}
```

`out.doc` comes from `frappe.client.get` which returns all standard fields so they leak into `fieldname`.

## Steps to reproduce

- load a doc with `createDocumentResource`
- change one field
- call `doc.save.submit()`

## Example

Payload sent by `save`

```json
{
  "doctype": "CRM Deal",
  "name": "CRM-DEAL-2026-00026",
  "fieldname": {
    "owner": "Administrator",
    "creation": "2026-06-26 17:25:10",
    "modified": "2026-06-26 17:25:10",
    "docstatus": 0,
    "idx": 0,
    "status": "Proposal/Quotation"
  }
}
```

`setValue` with a single field works since it does not include standard fields

```js
doc.setValue.submit({ status: 'Proposal/Quotation' }) // ok
```

## Expected

`save` should strip all standard fields or send only changed fields vs `originalDoc`

## Suggested fix

```js
makeParams() {
  const STD = ['doctype','name','owner','creation','modified','modified_by','docstatus','idx','parent','parentfield','parenttype']
  let values = JSON.parse(JSON.stringify(out.doc))
  STD.forEach((f) => delete values[f])
  return { doctype: out.doctype, name: out.name, fieldname: values }
}
```

ps173:
## Problem

`save.makeParams` serialized the entire doc and stripped only `doctype` and `name`, so read only standard fields returned by `frappe.client.get` (`owner`, `creation`, `modified`, `modified_by`, `docstatus`, `idx`) leaked into the `set_value` payload. The server rejected it with `Cannot edit standard fields` and the edit was never persisted.

## Fix

Diff `out.doc` against `out.originalDoc` and send only the changed fields. Since `originalDoc` is a snapshot taken right after load, unchanged standard fields never enter the payload. `doctype` and `name` are still stripped to cover the case where `originalDoc` is null.

## Tests

Added `src/resources/documentResource.test.ts` covering changed field only payloads, the no change case and the multi field subset case.

Fixes #802

Coverage: 58.03% (+0.05% vs `main`)

### DIFF
--- a/src/resources/documentResource.js
+++ b/src/resources/documentResource.js
@@ -108,7 +108,19 @@ export function createDocumentResource(options, vm) {
       {
         ...setValueOptions,
         makeParams() {
-          let values = JSON.parse(JSON.stringify(out.doc))
+          // send only the fields that changed vs the loaded doc, so that
+          // read-only standard fields (owner, creation, modified, docstatus,
+          // idx, ...) are not included in the set_value payload, which the
+          // server rejects with "Cannot edit standard fields"
+          let doc = JSON.parse(JSON.stringify(out.doc))
+          let originalDoc = out.originalDoc || {}
+          let values = {}
+          for (let key in doc) {
+            if (JSON.stringify(doc[key]) !== JSON.stringify(originalDoc[key])) {
+              values[key] = doc[key]
+            }
+          }
+          // never send identity/meta fields even if they appear changed
           delete values.doctype
           delete values.name
           return {
--- a/src/resources/documentResource.js
+++ b/src/resources/documentResource.js
@@ -108,25 +108,10 @@ export function createDocumentResource(options, vm) {
       {
         ...setValueOptions,
         makeParams() {
-          // send only the fields that changed vs the loaded doc, so that
-          // read-only standard fields (owner, creation, modified, docstatus,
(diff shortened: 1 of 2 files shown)


mehmehsloth:
Addressed the review nit in 367b3e9d

- save now short-circuits and resolves without a request when there are no pending changes, so it no longer fires set_value with an empty fieldname
- diff logic extracted into a shared getChangedFields helper used by both makeParams and the short-circuit check
- updated the no-change test to assert set_value is never hit

All 174 tests pass

PR #809, opened 2026-06-29
feat(MultiSelect): add creatable support for entering values not in options

ps173:
When `creatable` is set and the typed query matches no existing option, the popover shows a "Create <query>" row. Selecting it by click or by pressing Enter emits a new `create` event with the query string.

The component never adds the value to its own selection — the host listens for `create`, validates the query, and pushes it into v-model (and into the options list so the new entry resolves). The model value stays string[].

A sentinel value backs the create row and is prevented from ever entering modelValue: selecting the row calls preventDefault so the underlying listbox can't toggle it in, and an extra guard drops it on model updates as a safety net. The popover stays open after creating so users can add several entries in a row.

Includes a creatable story with removable chips, docs, and component tests covering the create row, click/Enter emits, and the sentinel guard.

Coverage: 58.89% (+0.03% vs `main`)


mehmehsloth:
Thanks for the review. Addressed the two actionable points in c0983fa2, and some clarification on the rest:

**Double-emit — verified it does *not* happen.** The concern is real in theory: the input is a reka `ComboboxInput`, so reka's Enter handler (`onKeydownEnter`) clicks the highlighted item, which would fire the create row's `@select` → `commitCreate`, while our own `@keydown.enter` also calls `commitCreate`. Whether both fire hinges on whether the create row is auto-highlighted, which depends on runtime timing (we pass `:ignore-filter="true"`), so I settled it empirically with a `calledOnce` assertion:

- Typical `type "mango" {enter}`: emits **once** (stable across repeated runs). The create row isn't auto-highlighted, so reka's Enter path no-ops and only our handler emits.
- Worst case — arrow-down to highlight the row, *then* Enter: also emits **once**. After the first `commitCreate`, `query` is cleared and `showCreateOption` flips `false` synchronously, so the `if (!showCreateOption.value) return` guard in `handleInputEnter` blocks the second emit.

I kept a `calledOnce` assertion on the Enter test to lock this in.

**Empty `create` description in `MultiSelect.api.md` — fixed.** `vue-component-meta` doesn't extract JSDoc for tuple-style emits in this repo (no plain emit anywhere gets a JSDoc-derived description — that's why `getEventDescription` keeps a hardcoded fallback map for `item-click`, `value-commit`, etc.). Added `create` to that map and regenerated, so the description now renders.

**On naming / "third way to create" (`creatable` vs Combobox's `allowCustomValue` vs `typeable`):** worth noting `MultiSelect` already uses reka `Combobox` internally — this isn't a parallel implementation. We're just surfacing the create-row affordance reka's combobox already supports and extending it with a host-driven `@create` contract (model stays `string[]`, host validates and commits). So the underlying primitive is the same; the open question is purely which *prop name* the frappe-ui family standardizes on. Happy to rename to whatever you and the team settle on — flagging for a maintainer call.

ssiyad:
@greptileai review

netchampfaris:
/barista review

Is it possible to have "create new" functionality using existing MultiSelect props and slots without introducing a new prop?

netchampfaris:
@claude Is it possible to have "create new" functionality using existing MultiSelect props and slots without introducing a new prop?

netchampfaris:
Status from the 1.0.0 selection sweep ([#871](https://github.com/frappe/frappe-ui/issues/871)): this is deferred past the `1.0.0` tag, not closed. It's additive, and additions freeze exactly like renames under [ADR-0012](https://github.com/frappe/frappe-ui/blob/main/spec/adr/0012-imperative-api.md), so it can land in a `1.x` minor once it meets [`spec/at-bar.md`](https://github.com/frappe/frappe-ui/blob/main/spec/at-bar.md) and gets the additive sign-off. Nothing in the freeze blocks it.

PR #818, opened 2026-07-02
feat: direct-mode pulse telemetry client

nextchamp-saqib:
Reworks the telemetry plugin to load the canonical pulse client from pulse's CDN
and post events directly to the ingest service — decoupled from the host backend's
framework version. Config comes from the whitelisted `boot_config` endpoint and
degrades to off when unreachable (old backends, network failure).

- Loads the client from the CDN; degrades to null when the asset can't be loaded
- `getDistinctId()` exposed for the anon → identity alias-on-signup flow
- Opt-in pageview tracking: pass your `vue-router` to auto-capture `pageview`
  per navigation, gated to new sites (`site_age <= 15`, like desk). Defaults to
  the matched route pattern (e.g. `/orders/:id`) so params never leave the browser
- `anonymousMode` (cookieless/client) plumbed through

All non-breaking: new options are optional, pageviews only activate when a router
is passed, and the router import is type-only (no runtime vue-router coupling).

Depends on frappe/frappe#40520 for the `site_age` gate.

Coverage: 68.65% (+9.80% vs `main`)


netchampfaris:
@nextchamp-saqib this should now be moved to framework?

PR #824, opened 2026-07-05
feat(editor): markdown format for useEditor and <Editor>

NagariaHussain:
## Why?

Frappe Wiki (see frappe/wiki#695) stores content as markdown. To use the new editor it currently has to hand-wire `@tiptap/markdown` around its own tiptap instance instead of using `useEditor`/`<Editor>`, because `format` only supports `'html' | 'json'`. Any markdown-backed app (wikis, docs, README editing) hits the same wall.

## What?

`format: 'markdown'` for `useEditor` and `<Editor>` — the content `v-model` carries a markdown string:

```vue
<Editor v-model="markdown" format="markdown" :extensions="[RichTextKit]" />
```

- Injects the `Markdown` extension from `@tiptap/markdown` automatically; configurable via the new `markdownOptions` option/prop (marked config, indentation)
- A caller-provided `markdown` extension wins over the injected one
- `Markdown` is re-exported from `frappe-ui/editor` for direct use
- Standard nodes/marks round-trip out of the box; custom extensions opt in with `renderMarkdown`/`parseMarkdown` (a follow-up can add these to frappe-ui's own nodes like Image/Video)

## How?

Same shape as the existing json path: `contentType: 'markdown'` at construction so initial content and external `setContent` parse as markdown, `getMarkdown()` in `onUpdate`/`change`, and the external-write guard extended with a markdown equality check. Docs + spec updated, unit tests added (injection, caller-override, bidirectional binding without echo).

https://claude.ai/code/session_01BjihfYwEXCQrwcuJ3yWLNM

Coverage: 68.82% (+0.01% vs `main`)

### DIFF
--- a/docs/content/docs/molecules/editor.md
+++ b/docs/content/docs/molecules/editor.md
@@ -28,6 +28,42 @@ The examples below are the common recipes. See [Exports](#exports) for the full 
 
 <ComponentPreview name="Editor-Inline" csr="true" />
 
+## Markdown editing
+
+For markdown-backed content (wikis, docs sites, README editors), set `format="markdown"` — the `v-model` then carries a markdown string. The editor parses it into the document on the way in and serializes back to markdown on every update, so the rest of your app never touches HTML.
+
+```vue
+<template>
+  <Editor v-model="markdown" format="markdown" :extensions="[RichTextKit]">
+    <template #default>
+      <EditorContent />
+    </template>
+  </Editor>
+</template>
+
+<script setup>
+import { ref } from 'vue'
+import { Editor, EditorContent, RichTextKit } from 'frappe-ui/editor'
+
+const markdown = ref('# Hello\n\nSome **markdown** content.')
+</script>
+```
+
+`format="markdown"` injects the `Markdown` extension from `@tiptap/markdown` automatically. Tune it with `markdown-options` (marked config, list/code indentation):
+
+```vue
+<Editor
+  v-model="markdown"
+  format="markdown"
+  :markdown-options="{ markedOptions: { breaks: true } }"
+  :extensions="extensions"
+/>
+```
+
+If your extension list already registers a `markdown` extension (for example an app-level `Markdown.configure(...)` — re-exported from `frappe-ui/editor`), that one wins and nothing is injected.
+
+Standard nodes and marks (headings, lists, tables, task lists, code blocks, links, images) round-trip out of the box. Custom nodes participate by defining `renderMarkdown`/`parseMarkdown` in their extension — nodes without them fall back to `@tiptap/markdown`'s defaults.
+
 ## Composing primitives
 
 When no kit-and-slot recipe fits, drop to `useEditor` and render the primitives yourself. This owns the editor instance directly — no `<Editor>` wrapper — and reads editor state (here, a live word count).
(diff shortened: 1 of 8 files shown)


netchampfaris:
@NagariaHussain Every consumer of useEditor will inherit the markdown bundle, can we make it optional? 

```html
import { Markdown } from 'frappe-ui/editor'

<Editor format="markdown" :extensions=[..., Markdown] />
```

This way markdown bundle is loaded only when it is imported.

yarn.lock
```diff
@@ -1462,6 +1462,13 @@
   resolved "https://registry.yarnpkg.com/@tiptap/extensions/-/extensions-3.26.0.tgz#0b3445bfdc37e8c307b3c0d23c117e9bf89ace27"
   integrity sha512-4wajuqnO2X0+LVvsBjW/xk3/tmdb16bNL939QhicAay4YYqXITeV2v3XJsryzmG4L5GkK1yLxvRGk4aLoxWrnA==
 
+"@tiptap/markdown@^3.26.0":
+  version "3.27.1"
```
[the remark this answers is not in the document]
NagariaHussain: Fixed in a37b1dd3 — lockfile now resolves @tiptap/markdown to 3.26.0, matching the rest of the tiptap family (3.26.0's peer deps pin core/pm to exactly 3.26.0).

src/molecules/editor/useEditor.ts
```diff
@@ -19,7 +19,12 @@ type Editor = TiptapEditor
 
 export type UseEditorOptions = {
   content?: Ref<string | JSONContent | null | undefined>
-  format?: 'html' | 'json'
+  /**
+   * `'markdown'` requires the `Markdown` extension (re-exported from
+   * `frappe-ui/editor`) in `extensions` — kept explicit so only editors that
+   * use markdown pay for its bundle.
+   */
```
NagariaHussain: Not needed.

src/molecules/editor/useEditor.ts
```diff
@@ -88,6 +109,12 @@ export function useEditor(
     },
   }
 
+  if (format === 'markdown') {
+    // Makes the initial `content` (and paste-adjacent internals) parse as
+    // markdown instead of being sniffed as HTML/JSON.
```
NagariaHussain: Remove this comment too

NagariaHussain:
@netchampfaris done.

src/molecules/editor/useEditor.ts
NagariaHussain: Removed in be5bae70.

src/molecules/editor/useEditor.ts
NagariaHussain: Removed in be5bae70.

docs/content/docs/molecules/editor.md
```diff
@@ -28,6 +28,35 @@ The examples below are the common recipes. See [Exports](#exports) for the full
...
+
+<script setup>
+import { ref } from 'vue'
+import { Editor, EditorContent, RichTextKit, Markdown } from 'frappe-ui/editor'
+
+const markdown = ref('# Hello\n\nSome **markdown** content.')
+</script>
+```
+
+The `Markdown` extension is passed explicitly (rather than injected by the prop) so its parser/serializer only lands in bundles of editors that use it. Tune it with `Markdown.configure` (marked config, list/code indentation):
```
NagariaHussain: Remove () note

NagariaHussain: Also bundle note is not needed here public docs this is

NagariaHussain: Removed both — comment now just points at Markdown.configure.

PR #827, opened 2026-07-06
fix(calendar): dark-mode event colors

BreadGenie:
Docs preview: https://ui.frappe.io/pr-preview/pr-827/docs/components/calendar

Coverage: 68.75% (-0.05% vs `main`)


BreadGenie:
I'm merging this for now so that the calendar themes aren't broken. We can clean this up regardless when we do the calendar rewrite.

PR #828, opened 2026-07-06
feat(avatar): introduce themes for avatar

BreadGenie:
defaults to gray, so no breaking change, but let's you override with few themes similar to Badge and has auto option that picks a deterministic value based on the label.

Coverage: 68.83% (+0.01% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-828/

### DIFF
--- a/src/components/Avatar/Avatar.md
+++ b/src/components/Avatar/Avatar.md
@@ -6,10 +6,13 @@ A visual representation of a user, typically shown as an image, initials, or an 
 
 <ClientOnly><AvatarBuilder /></ClientOnly>
 
-## Themes
+## Shapes
 <ComponentPreview name="Avatar-Shapes" />
 
-## Controlled state
+## Sizes
 <ComponentPreview name="Avatar-Sizes" />
+
+## Themes
+<ComponentPreview name="Avatar-Themes" />
 
 <!-- @include: ./Avatar.api.md -->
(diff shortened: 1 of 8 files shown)


src/components/Avatar/Avatar.md
```diff
@@ -6,10 +6,13 @@ A visual representation of a user, typically shown as an image, initials, or an
...
-## Themes
+## Shapes
 <ComponentPreview name="Avatar-Shapes" />
 
-## Controlled state
+## Sizes
 <ComponentPreview name="Avatar-Sizes" />
 
+## Themes
+<ComponentPreview name="Avatar-Themes" />
```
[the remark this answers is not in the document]
BreadGenie: added the missing file

src/components/Avatar/types.ts
```diff
@@ -1,3 +1,13 @@
+export type AvatarTheme =
+  | 'gray'
+  | 'blue'
+  | 'green'
+  | 'amber'
+  | 'red'
+  | 'violet'
+  | 'orange'
+  | 'auto'
```
netchampfaris: @BreadGenie if Avatar has theme=auto, all other components that have theme prop will also expect theme=auto

BreadGenie: ig the "auto" theme computation should be done in app instead?

BreadGenie: removed auto as well as orange (marked depracated in badge component) since I noticed other components seem to be using the new amber option instead of orange

PR #829, opened 2026-07-07
feat(SettingsDialog): add close button

netchampfaris:
## Problem

`SettingsDialog` renders the base `Dialog` in `bare` mode, and in bare mode the base Dialog deliberately skips its own close button. That left the settings dialog with **no visible close affordance** — users could only dismiss it via Esc or a backdrop click, which isn't discoverable.

## Change

Add a floating ghost X button in the dialog's top-right corner. It closes the dialog by setting the `v-model`, so route-driven consumers (e.g. Gameplan, whose open state is derived from the URL) still react through their own open-state watcher — no consumer changes needed.

- Made the `TabsRoot` container `relative` to anchor the button.
- Button styling mirrors the standalone close button already used by the base `Dialog` (ghost variant, `lucide-x`, `text-ink-gray-9`).

## Verification

Verified in Gameplan against the local demo site: opening Settings shows the X in the top-right (no overlap with panel header content), and clicking it closes the dialog and navigates back to the underlying page.

Docs preview: https://ui.frappe.io/pr-preview/pr-829/

Coverage: 68.81% (+0.01% vs `main`)

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -12,8 +12,21 @@
       orientation="vertical"
       activation-mode="manual"
       :unmount-on-hide="unmountOnHide"
-      class="flex h-[100dvh] min-h-0 w-screen flex-col overflow-hidden sm:h-[min(860px,calc(100vh-8rem))] sm:min-h-[560px] sm:w-auto sm:flex-row"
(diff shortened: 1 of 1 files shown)


netchampfaris:
Discussed with Timeless, in most cases there is a primary action at the top right so close button position will clash. Clicking outside the dialog or pressing escape is the way to close. If anyone needs a close button, They can add it explicitly on the main slot.

PR #830, opened 2026-07-10
fix(Sidebar): center item icons when collapsed

NagariaHussain:
## Problem

When the sidebar is collapsed, item icons sit hard left in the rail instead of centered, and the active item's pill shows a large empty area to the right of the icon.

The label span keeps `flex-1` in the collapsed state. Its `flex-basis: 0%` takes precedence over `w-0`, so the grow factor expands the invisible label to fill the row — `justify-center` on the link has no free space left to center in.

## Solution

Apply `flex-none` to the label span when collapsed (and `flex-1` only when expanded), so the hidden label truly collapses to zero width and the icon centers in the rail.

| Before | After |
|--------|-------|
| Icon left-hugged, empty pill space right | Icon centered, compact pill |

https://claude.ai/code/session_012PXJnZXUyt5vyE7DiuLnD3

Coverage: 68.82% (±0.00% vs `main`)

### DIFF
--- a/src/components/Sidebar/SidebarItem.vue
+++ b/src/components/Sidebar/SidebarItem.vue
@@ -44,12 +44,15 @@
           </slot>
         </span>
       </Tooltip>
+      <!-- flex-none when collapsed: with flex-1, the grow factor would override
+           w-0 (flex-basis 0% wins over width) and the invisible label would
+           still fill the row, pushing the icon off-center. -->
       <span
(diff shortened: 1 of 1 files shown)


netchampfaris:
@NagariaHussain screenshots?

NagariaHussain:
It is not ready for review yet, hence in draft 😅

netchampfaris:
@NagariaHussain open when ready

PR #831, opened 2026-07-11
fix: bump reka-ui to 2.9.9 so dialog inputs focus on mouse click

### LINKED ISSUE #766: Dialog: mouse clicks can't focus inputs (overlay pointerdown.preventDefault suppresses mousedown)
### Summary

In a `Dialog`, **mouse clicks cannot focus text inputs/textareas/contenteditable**. Keyboard `Tab` + typing works, and `click`-driven controls (Close/Submit buttons) work — only fields that rely on mouse focus are dead. Observed on the `v1/espresso-tokens` branch.

### Root cause

`Dialog.vue` uses the scrollable layout where `<DialogContent>` is nested **inside** a scrollable `<DialogOverlay>` (`overflow-y-auto`) — the standard pattern for tall, scrollable dialogs.

However, the overlay element ends up with a `pointerdown` handler bound through Vue's `.prevent` modifier (confirmed live: the canceling listener's `currentTarget` is the overlay div, owner `Primitive`). Because content is nested under the overlay, a `pointerdown` on a dialog input **bubbles up to the overlay and is `preventDefault()`'d**.

Per the Pointer Events spec, cancelling `pointerdown` suppresses the compatibility `mousedown` — and `mousedown`'s default action is what focuses an input + places the caret. Net effect:

- `pointerdown` and `click` fire on the input, but **`mousedown` and `focusin` never do**.
- Inputs/textareas/contenteditable can't be focused by mouse.
- Buttons still work because they act on `click`.

Note: reproduces **only with a trusted (real) mouse event**. Synthetic `el.click()` / `dispatchEvent` bypasses the focus-on-mousedown machinery and will *not* reproduce it.

### Open question

I could not locate where the overlay's `pointerdown.prevent` is bound — it is not in `Dialog.vue`'s template, nor anywhere in reka-ui's `Dialog/*` source. It arrives on the overlay element via the `$attrs` forwarding chain (`DialogOverlay` → `DialogOverlayImpl` → `Primitive`). Tracking down and removing it at the source is the proper fix.

### Current workaround (shipped)

`@pointerdown.stop` on `<DialogContent>` so content pointerdowns never reach the overlay's prevent handler. Verified: input focus, typing, Escape, buttons, and outside-click dismiss all work (reka's dismiss detection is capture-phase, unaffected by a bubble-phase `.stop`).

### Proper fix options

1. Find and remove/scope the stray `pointerdown.prevent` on the overlay (ideal — root cause).
2. Keep the workaround `@pointerdown.stop`.
3. Restructure to reka's centered sibling layout (`DialogOverlay`/`DialogContent` as siblings, content `fixed`-centered) — note this trades away in-overlay scrolling for tall dialogs, and a naive sibling scroll-wrapper would render even when closed (`DialogPortal` is a bare `Teleport`), so it needs its own `Presence` gating.

mvanhorn:
## Summary
Mouse clicks can once again focus inputs, textareas, and contenteditable regions inside a `Dialog`. reka-ui 2.9.8 shipped a regression where the dialog overlay cancelled `pointerdown` events bubbling up from dialog content, which suppressed the compatibility `mousedown` that performs focus and caret placement — so keyboard focus worked but mouse focus was dead.

## Why this matters
Reported in #766. The component nests `<DialogContent>` inside a scrollable `<DialogOverlay>` — an upstream-supported pattern for scrollable dialogs. reka-ui 2.9.8 (unovue/reka-ui#2655) added `@pointerdown.left.prevent` to `DialogOverlayImpl` to restore trigger focus after backdrop clicks, but it also cancelled pointerdowns bubbling from nested content. This was fixed upstream in 2.9.9 (unovue/reka-ui#2668) by scoping the prevention with `.self`. Only 2.9.8 was affected — frappe-ui `main`'s own lockfile resolved 2.7.0 (unaffected) — but the `^2.5.0` floor let consumer lockfiles resolve the broken 2.9.8, which is how #766 was hit.

## Changes
- Bumps the `reka-ui` dependency floor to `^2.9.9` so no consumer can resolve into the broken 2.9.8, and updates the lockfile accordingly.
- Removes the `@pointerdown.stop` workaround on `<DialogContent>`, obsolete now that the root cause is fixed upstream.
- Rewrites the dismissible dialog tests with a stateful wrapper (the previous props-only mounts could never close, making the assertions vacuous) and adds outside-click dismiss and scrollable-dialog focus coverage.

## Testing
`yarn install` resolves reka-ui to 2.9.9; verified a dialog with a text input focuses on a real mouse click (caret placed) as well as via keyboard. Full CI is green, including all Cypress component-test shards.

Fixes #766

Coverage: 68.75% (-0.03% vs `main`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,7 +193,7 @@
     "ora": "5.4.1",
     "prettier": "^3.3.2",
     "prosemirror-tables": "^1.8.1",
-    "reka-ui": "^2.5.0",
+    "reka-ui": "^2.9.9",
     "slugify": "^1.6.6",
     "socket.io-client": "^4.5.1",
     "tippy.js": "^6.3.7",
--- a/src/components/Dialog/Dialog.cy.ts
+++ b/src/components/Dialog/Dialog.cy.ts
@@ -86,19 +86,90 @@ describe('Dialog', () => {
 
   // ---- New behavior props ----------------------------------------------------
 
-  it('dismissible=false blocks outside click and Escape from closing', () => {
-    cy.mount(Dialog, {
-      props: {
-        open: true,
-        title: 'Locked',
-        message: 'Cannot dismiss',
-        dismissible: false,
+  it('closes on outside click when dismissible', () => {
+    const Wrapper = defineComponent({
+      setup() {
+        const open = ref(true)
+        return { open }
+      },
+      render() {
+        return h(Dialog, {
+          open: this.open,
+          'onUpdate:open': (v: boolean) => (this.open = v),
+          title: 'Dismissible',
+        })
       },
     })
 
+    cy.mount(Wrapper)
+    cy.get('[role=dialog]').should('exist')
+    cy.get('.dialog-overlay').click(0, 0, { force: true })
+    cy.get('[role=dialog]').should('not.exist')
+  })
+
+  it('dismissible=false blocks outside click and Escape from closing', () => {
+    const Wrapper = defineComponent({
+      setup() {
+        const open = ref(true)
+        return { open }
+      },
+      render() {
+        return h(Dialog, {
+          open: this.open,
+          'onUpdate:open': (v: boolean) => (this.open = v),
+          title: 'Locked',
+          message: 'Cannot dismiss',
+          dismissible: false,
+        })
+      },
+    })
+
+    cy.mount(Wrapper)
+    cy.get('[role=dialog]').should('exist')
+    cy.get('.dialog-overlay').click(0, 0, { force: true })
     cy.get('[role=dialog]').should('exist')
     cy.get('body').type('{esc}')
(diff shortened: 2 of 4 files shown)


mvanhorn:
Thanks @netchampfaris! The reka-ui bump fixing dialog input focus on mouse click was a satisfying one to land.

netchampfaris:
@mvanhorn Hey, how are you using frappe-ui in your projects?

mvanhorn:
Honestly I'm not building on it day to day - I keep an eye on interesting projects and frappe-ui was one of them, so when I saw dialog inputs losing mouse focus after the reka-ui 2.9.8 bump I wanted to dig in and contribute the fix. Enjoyed poking around the codebase.

PR #836, opened 2026-07-13
Optimize first recipe with eager iframe loading for faster above-the-fold rendering

netchampfaris:
## Summary
This PR optimizes the initial page load performance by enabling eager iframe loading for the first recipe component, which appears above the fold. Instead of waiting for the Intersection Observer to detect viewport visibility, the first recipe now renders its iframe directly into the server HTML, allowing the browser to begin fetching the demo before Vue hydration completes.

## Key Changes
- Added `eager` prop to `RecipeExample` component to control iframe loading strategy
- Modified iframe visibility initialization to respect the `eager` flag, starting visible immediately for eager recipes
- Implemented dynamic `loading` attribute on iframes: uses `"eager"` for the first recipe and `"lazy"` for others
- Updated Intersection Observer setup to skip initialization when the recipe is already visible (eager case)
- Applied the `eager` flag to the first recipe in the documentation homepage
- Enhanced code comments to explain the eager loading optimization and its rationale

## Implementation Details
- The `visible` ref now initializes to `props.eager ?? false` instead of always `false`
- A computed property `iframeLoading` determines the appropriate loading strategy based on the `eager` prop
- The observer setup includes an early return when `visible.value` is true, preventing unnecessary observer attachment for eager recipes
- Both desktop and mobile iframe elements now use the dynamic `:loading` binding instead of the static `loading="lazy"` attribute

https://claude.ai/code/session_01C5Dkri6xLianhSoB8oqhUf

Docs preview: https://ui.frappe.io/pr-preview/pr-836/

Coverage: 68.78% (+0.03% vs `main`)

### DIFF
--- a/docs/components/recipes/RecipeExample.vue
+++ b/docs/components/recipes/RecipeExample.vue
@@ -5,7 +5,9 @@
 // slot (`#desktop` / `#mobile`) from a `<<<` include so VitePress highlights it.
 //
 // The iframe src is only set once the block scrolls near the viewport, so the
-// page doesn't boot every app shell at once.
+// page doesn't boot every app shell at once. The exception is the first recipe
+// (`eager`), which sits in the first fold: it renders its iframe straight into
+// the server HTML so the browser fetches the demo before Vue even hydrates.
 import { computed, onBeforeUnmount, onMounted, ref, useTemplateRef } from 'vue'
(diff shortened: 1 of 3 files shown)


netchampfaris:
Not much difference, not merging yet

PR #837, opened 2026-07-14
fix(editor): mount suggestion popup via renderer wrapper

RitvikSardana:
This PR contains two fixes for the editor suggestion popups: the popup not mounting in production builds, and `SuggestionExtension` dropping the `allowSpaces` option.

### Problem

The `@` mention and custom `{{` suggestion popups silently fail to open in **production builds** (they work in dev). No error is thrown; the popup component renders but is never attached to the DOM.

### Root cause

`createSuggestionRenderer` (`molecules/editor/extensions/shared/suggestion-renderer.ts`) attaches `renderer.element` to `document.body`. In `@tiptap/vue-3`'s `VueRenderer`, `element` is the mount wrapper's **`firstElementChild`**, computed once at construction.

`SuggestionList` renders through `EditorPopover` → reka-ui `FocusScope`. With **reka-ui >= 2.5**, `FocusScope` no longer exposes a synchronous first element child at render time, so `renderer.element` is `null` when `onStart` runs → the early-return guard fires and the popup is never appended. `onUpdate` only repositions, so it never recovers.

This is timing/reka-version dependent, which is why it reproduces only in prod builds (and why older reka, e.g. gameplan on reka 2.0.x, isn't affected).

### Fix

Attach and position the `VueRenderer` **wrapper** (`renderer.el`) instead of its `firstElementChild`. The wrapper always exists right after `new VueRenderer(...)`, and the popup renders inside it, so this is robust regardless of the popup component's root element or the reka-ui version. A late-attach fallback in `onUpdate` covers any case where the initial attach couldn't complete. The existing stale-query token / teardown guards are preserved.

### Problem

The legacy `createSuggestionExtension` factory supported `allowSpaces`, which keeps the suggestion popup open when the query contains a space (e.g. typing `{{first response` to match a "First Response Time" field). The new `SuggestionExtension.configure` wrapper builds the tiptap suggestion config from a fixed set of keys and silently drops everything else, so consumers migrating to it lose this behavior: the popup dismisses the moment a space is typed.

### Fix

Add `allowSpaces?: boolean` to `SuggestionExtensionOptions` and forward it to the underlying `@tiptap/suggestion` plugin config, restoring parity with `createSuggestionExtension`. Covered by a unit test in `useEditor.test.ts`.

## How to test

1. Build an app that uses `RichTextKit` mentions or a `SuggestionExtension` (e.g. Frappe Helpdesk) in **production mode** with reka-ui >= 2.5.
2. In the rich-text editor, type `@` (mention) or a custom trigger like `{{`.
3. Before: no popup appears. After: the suggestion list opens and filters as you type.
4. For `allowSpaces`: configure a `SuggestionExtension` with `allowSpaces: true`, type a multi-word query like `{{first response`. Before: the popup closes on the space. After: it stays open and keeps filtering.

Verified in a Helpdesk production build for both `@` mentions and `{{` field autocomplete, and in dev for `allowSpaces` with the Saved Replies `{{` field autocomplete.

Coverage: 68.85% (+0.10% vs `main`)

### DIFF
--- a/src/molecules/editor/SuggestionExtension.ts
+++ b/src/molecules/editor/SuggestionExtension.ts
@@ -16,6 +16,7 @@ export type SuggestionExtensionOptions<TItem = any> = {
   items: TItem[] | ((query: string) => TItem[] | Promise<TItem[]>)
   component?: Component
   floatingOptions?: SuggestionFloatingOptions
+  allowSpaces?: boolean
   command: (props: { editor: Editor; item: TItem; range: SuggestionRange }) => void
 }
 
@@ -26,6 +27,7 @@ function buildSuggestionExtension<TItem = any>(options: SuggestionExtensionOptio
       return {
         suggestion: {
           char: options.trigger,
+          allowSpaces: options.allowSpaces,
           pluginKey: new PluginKey(options.name),
           items: ({ query }: { query: string }) =>
             typeof options.items === 'function' ? options.items(query) : options.items,
--- a/src/molecules/editor/extensions/shared/suggestion-renderer.test.ts
+++ b/src/molecules/editor/extensions/shared/suggestion-renderer.test.ts
@@ -118,6 +118,20 @@ describe('createSuggestionRenderer', () => {
     expect(document.body.contains(renderer.el)).toBe(true)
   })
 
+  it('keeps previous items during a transient loading update', () => {
+    const api = createSuggestionRenderer(FakeComponent)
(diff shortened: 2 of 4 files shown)


RitvikSardana:
@netchampfaris let us merge this?

Why it works in CRM as of now? 
CRM has not migrated to frappe-ui's new Editor


Why it works in Gameplan?
because it has  `"reka-ui": "^2.0.2"` in package.json,  and in frappe-ui the version is `"reka-ui": "^2.9.9",`

PR #839, opened 2026-07-14
feat(Progress): animate the continuous fill on value change

### LINKED ISSUE #838: Progress: fill jumps between value updates — add a width transition for smooth animation
## Describe the bug

The continuous `Progress` bar has no transition on its fill, so every `value` update makes the bar jump to the new width in a single frame. When the value is driven by periodic updates (polling a job, a provisioning heartbeat, simulated progress), the bar advances in visible steps and feels janky instead of gliding smoothly between values.

Preview of issue:

https://github.com/user-attachments/assets/34bd5bc3-dc3e-40bb-b582-406aae480529

The fill div renders the width directly with no easing:

https://github.com/frappe/frappe-ui/blob/main/src/components/Progress/Progress.vue#L30-L35

```vue
<!-- Continuous Progress Bar -->
<div
  v-if="!props.intervals"
  class="h-full bg-surface-gray-10"
  :style="`width: ${props.value}%`"
></div>
```

## Steps to reproduce

1. Render `<Progress :value="value" size="lg" />`
2. Advance `value` on a timer, e.g. `setInterval(() => (value += 8), 500)` — or anything that mimics real progress reporting
3. The bar snaps forward on each tick instead of animating toward the new width

## Expected behavior

The fill should ease toward the new width so discrete updates read as one continuous motion — e.g. a `transition-[width] duration-300 ease-out` (ideally with a duration that comfortably covers typical update intervals). For reference, this is how we ended up styling a hand-rolled bar to get the feel we wanted:

```html
<div class="h-full rounded-full bg-surface-gray-10 transition-[width] duration-700 ease-out" :style="{ width: `${progress}%` }" />
```

It would be nice to have this built in — possibly with the duration configurable via a prop, and disabled under `prefers-reduced-motion`.

## Context

- frappe-ui `1.0.0-beta.24`
- Noticed while building an onboarding/provisioning screen where progress creeps forward every few hundred ms; the stepping is very noticeable on `size="lg"`.

vinaypokharkar:
The continuous `Progress` fill set `width` inline with no transition, so every
value change snapped in a single frame. Apps that drive progress from a timer or a
poll rendered a bar that jumped in discrete steps.

The fill now transitions, using the same duration, curve and technique as
[Radix UI's progress demo](https://github.com/radix-ui/website/blob/main/components/demos/progress/css/styles.css):
660ms `cubic-bezier(0.65, 0, 0.35, 1)`, sliding a full-width fill with
`translateX` instead of animating `width`. `transform-gpu` on the track is part
of that — without a compositing layer Safari fails to clip the animating fill to
the rounded corners.

Also rebuilt on reka-ui's `ProgressRoot` / `ProgressIndicator`, replacing the
hand-rolled div and its manual ARIA. Reka supplies `role`, `aria-valuemin/max/now`
and `data-state` / `data-value` / `data-max`.

The interval bar is untouched — it swaps background colour rather than width, so
it never animated, and a test asserts it stays that way.

Fixes #838

## Notes for reviewers

- **Default-on for downstream apps.** Every consumer gets 660ms of animation on
  existing `Progress` usages with no code change. There's no per-call opt-out;
  `prefers-reduced-motion` is the only escape hatch (`motion-reduce:transition-none`).
- **Out-of-range values now clamp.** `width: -20%` used to resolve to nothing.
  Both ends are pinned and tested — reka also rejects an out-of-range value and
  falls back to indeterminate, so the clamp happens before it.
- **Cosmetic edges, not addressed:** resetting `value` to 0 animates the bar
  backwards, and the fill reaches 100% ~660ms after the value does. Both are
  inherent to animating toward a target.
- The docs page was restructured around usage patterns (upload, storage quota,
  onboarding checklist, multi-step form) in place of one-prop-per-story examples.

## Testing

8 passing in `Progress.cy.ts`. The transition assertions read **computed** CSS
rather than class strings, so they fail if a utility stops resolving to a real
`transition-property`.

Coverage: 66.33% (+0.01% vs `main`)


netchampfaris:
/barista review

do we need to add duration prop? feel like its unnecessary

netchampfaris:
Pushed docs work plus component changes on top of this. Flagging what goes beyond the original scope, since some of it is deletions.

**Stories** — removed `Examples`, `Hint` and `Intervals`; each set one prop and rendered nothing else, which the playground above already does interactively. Added `Animated`, `Storage`, `Onboarding` and `MultiStepForm` from the espresso 2.0 usage examples, all checked in the browser. `Intervals` is the one worth a second opinion: it's real behaviour, so it's now shown in `MultiStepForm` doing the job it exists for rather than frozen at 30%. Easy to restore a minimal version.

**Component**

- Fill animates `translateX` instead of `width` — animating `width` reflows the bar every frame. Already noted as a follow-up here; small enough to just do.
- Duration and curve match [Radix's progress demo](https://github.com/radix-ui/website/blob/main/components/demos/progress/css/styles.css): 660ms `cubic-bezier(0.65, 0, 0.35, 1)`. This replaces your `ease-linear` reasoning — say the word if the slow start bothers you on chained updates.
- Rebuilt on reka-ui's `ProgressRoot` / `ProgressIndicator`, so `role`, `aria-value*` and `data-state` come from the primitive.
- Out-of-range values clamp now; `width: -20%` used to resolve to nothing. Both ends tested.

Description rewritten too — it still documented the `duration` prop and its playground knob.

PR #845, opened 2026-07-20
fix: mention dropdown stays dead after typing a name with no match

kaulith:
## Summary

Same issue as #844, but in `frappe-ui/editor`.

Mention suggestions could get stuck after typing a query with no matches, then deleting back to a matching prefix. For example, typing `@farxyz`, pressing space, then deleting back to `@far` never restored the dropdown, even though matching users existed.

This happened because `SuggestionList.vue` conditionally rendered its root element. When the root switched between a component and a comment node, the detached popover content could no longer be updated, and if suggestions started while there were no matches, no popover was mounted at all.

## Changes

- Keep a stable wrapper element mounted at all times.
- Move the `v-if` to `EditorPopover`, ensuring the same DOM node is reused across renders.

## Result

Mention suggestions now recover correctly when the query changes from no matches back to matching results.

## Before

https://github.com/user-attachments/assets/2f9fc59a-d4a2-4674-b733-db7a6eb61a64

## After

https://github.com/user-attachments/assets/dbc39a13-b025-4836-a185-b8a1f07d9544

Coverage: 68.78% (+0.02% vs `main`)

### DIFF
--- a/src/molecules/editor/extensions/suggestion/SuggestionList.vue
+++ b/src/molecules/editor/extensions/suggestion/SuggestionList.vue
@@ -1,49 +1,51 @@
 <template>
-  <EditorPopover
-    v-if="items.length || showNoResults"
-    dialog-label="Suggestions"
-    :autofocus="false"
-    :trapped="false"
-    :loop="false"
-    :content-class="[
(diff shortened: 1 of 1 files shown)


kaulith:
@netchampfaris Thanks for letting me know, wasn't aware of editor.
Tried and found same issue here, please review 🫠

PR #856, opened 2026-07-24
feat(MultiSelect): add search prefix and suffix slots

### LINKED ISSUE #855: MultiSelect should expose a suffix slot on the search input
### Summary

`MultiSelect`'s **search input** row exposes no slot for trailing content. The search row currently renders only the `ComboboxInput` and the loading indicator:

```vue
<div v-if="!hideSearch" data-slot="search" class="flex items-center gap-2 border-b border-outline-gray-1 px-3">
  <ComboboxInput data-slot="input" ... />
  <LoadingIndicator v-if="loading" ... />
</div>
```

So there is no way to render something at the end of the search field, e.g.:

- a keyboard-shortcut key-cap (like the one shown in the search box for filters),
- a "clear query" button,
- a small filter/scope icon.

`TextInput` already supports `#prefix` / `#suffix` for exactly this; `MultiSelect`'s search field should offer the same.

### Proposal

Add a **`suffix`** slot inside the search row:

```vue
<div v-if="!hideSearch" data-slot="search" ...>
  <ComboboxInput data-slot="input" ... />
  <LoadingIndicator v-if="loading" ... />
  <slot name="suffix" v-bind="slotProps" />
</div>
```

Usage:

```vue
<MultiSelect :options="options">
  <template #suffix>
    <kbd>G</kbd>
  </template>
</MultiSelect>
```

### Note on naming

`MultiSelect` already has a `suffix` slot on the **default trigger** (the chevron area). That outlet only renders in the default-trigger branch (`v-else`), so it does not collide when a custom `#trigger` is supplied. Reusing the name `suffix` for the search field keeps the consumer API simple and matches `TextInput`, but if the maintainers prefer to avoid the overload, a search-scoped name (e.g. `search-suffix`) or a matching `#prefix` pair would also work. Flagging so the final name is a deliberate choice.

netchampfaris:
## Summary

- add `#search-prefix` and `#search-suffix` slots around MultiSelect's in-popover search input
- expose scoped query controls for reading, setting, clearing, and refocusing the search
- preserve the existing trigger `#prefix` / `#suffix` behavior
- export the new slot prop types and update generated API documentation
- add a documented story demonstrating a search icon, shortcut hint, and clear-query action

## Why

Consumers currently cannot add keycaps, clear-query controls, scope indicators, or similar affordances to the search row without replacing component internals. Namespaced search slots provide that extension point without colliding with the trigger suffix.

Closes #855.

## Validation

- `yarn test` — 202 tests passed
- `yarn cypress run --component --spec src/components/MultiSelect/MultiSelect.cy.ts` — 20 tests passed
- `node --max-old-space-size=4096 node_modules/vitepress/bin/vitepress.js build docs`
- `git diff --check`

Repository-wide `yarn type-check` remains blocked by existing errors outside the changed MultiSelect files.

Docs preview: https://ui.frappe.io/pr-preview/pr-856/

Coverage: 66.22% (+0.05% vs `main`)


netchampfaris:
Superseded by #858.

This PR started as two slots around MultiSelect's search box. Reviewing that one pair meant looking at the rest of the family, and the scope grew far past what this branch and title describe — it now freezes the whole public API of `Select`, `Combobox` and `MultiSelect` before 1.0.0, with breaking renames and removals.

Rather than leave a PR whose title and description no longer match its contents, the work continues on a renamed branch in #858. The search slots are still in there, on both Combobox and MultiSelect. #858 carries "Closes #855".

Closing this one. Review history stays here for the record.

Issue #870, opened 2026-07-30
Root-surface audit: non-component exports from src/index.ts


netchampfaris:
## Question

What is leaking out of `src/index.ts` that should not freeze at `1.0.0`?

The sweep tickets cover components. This one covers everything else the root export hands out — utilities, composables, types, directives, and plugins — plus anything exported by accident.

`src/utils/` holds 26 modules and `src/composables/` holds 11. Some are genuinely public API (`useTheme`, `toast`, `dialog`, `useScreenSize`). Some are internal plumbing that happens to be reachable. Once `1.0.0` is tagged, both categories are frozen equally.

Produce, for every non-component export from `src/index.ts`:

- is it intentional public API, or an accident?
- if intentional: does its name and shape match `PHILOSOPHY.md` and `CONTEXT.md` vocabulary, and is it documented?
- if accidental: un-export it before the tag
- do any of them belong on a subpath rather than root, under the rule from the subpath-export ticket?

Watch specifically for: `warnDeprecated`, `iconString`, `frappeRequest` / `request` / `call`, `socketio`, `fileUploadHandler` / `useFileUpload`, `pageMeta`, `dayjs`, `markdown`, `url-validation`, `useId`, `vnode`, `debounce`, and the `mocks/` directory. Also check `directives/` and any bare `export *` that pulls more than intended.

Blocked on the subpath rule, since "move it to a subpath" is only available once that rule exists.

netchampfaris:
## Resolution

**The root surface has no accidental-export problem where this ticket expected one, and `dayjs` has no cost problem at all.** Both premises died on contact with measurement. What is actually wrong is duplication, naming, and a `export *` habit — 17 members come off the surface, 12 stay, and 6 barrel lines get replaced.

Resolving all 434 root exports through the TypeScript compiler: **none** of the watch list leaks. `warnDeprecated`, `iconString`, `markdown`, `url-validation`, `useId`, `vnode` and `mocks/` are unreachable from an entry point. The hunt was aimed at the wrong modules.

### Scope

[...]

Issue #885, opened 2026-07-30
Charts and VueGridLayout: ship in 1.0.0 or un-export


netchampfaris:
## Question

Do `Charts` and `VueGridLayout` ship in `1.0.0`?

Both are exported today. `v1-release/plan.md` lists `VueGridLayout` as explicitly not core for v1 and its modernization as an explicit non-blocker; `Charts` was never classified at all.

Under this map's freeze rule, "not core but still exported" is not a resting place — anything exported at the tag is frozen until `2.0.0`. So each needs a real call:

- **ship it** — then it gets the full at-bar checklist like any other family, including a docs page that matches the code
- **un-export it** — with a changelog entry, a migration note, and a plan for whichever apps currently import it

Establish the facts before deciding:

- who actually consumes each one, across the bench apps
- what each depends on — `VueGridLayout` wraps a third-party grid library, and `Charts` may too. A heavy third-party dependency is also an argument for a subpath rather than root, so coordinate with the subpath rule.
- how far below bar each is (TypeScript, tests, docs page accuracy)

Un-exporting is the recommended default for anything with thin consumption and a heavy dependency — the library can always add it back in a minor, but it cannot remove it before `2.0.0`.

netchampfaris:
From #870: the bundle-cost premise here is measured false. A real consumer build (frappe-ui resolved from `node_modules`, vite 7 / rollup) shows `import { Button } from 'frappe-ui'` drops `echarts` and `grid-layout-plus` entirely — root imposes neither on a consumer who doesn't import the chart components. The keep-or-unexport question stands; the P15 limb-1 argument for it does not. Rolldown is untested (#903).

netchampfaris:
## Resolution

**Both families come off. The old `Charts` are deleted and replaced by charts v2 ([#890](https://github.com/frappe/frappe-ui/pull/890)) at `frappe-ui/charts`; `GridLayout` is deleted with no replacement.** Neither reaches bar, and the ticket's recommended default — un-export anything with thin consumption and a heavy dependency — survives contact with the numbers on both.

### Rule 5 recount: the ticket's own premise inverted, for the third time on this map

Counted against freshly fetched remote refs, with every import resolved to `frappe-ui` rather than counting tags:

| Export | Sites | Where |
| --- | --- | --- |
| `AxisChart` | 3 | crm, helpdesk, builder |
| `ECharts` | 3 | helpdesk only |
| `DonutChart` | 2 | crm, helpdesk |
[...]

PR #890, opened 2026-08-03
feat: charts v2

nextchamp-saqib:
A new chart library at `frappe-ui/charts`, built from scratch. It is meant to be the standard chart library for every Frappe app.

**Charts:** Bar, Line, Area, Scatter, Donut (full and half), Funnel, Heatmap, Sankey, NumberCard.

```vue
<BarChart :data="rows" x="month" :y="['revenue', 'costs']" stacked />
<LineChart :data="tidyRows" x="month" y="amount" series="region" />
```

Props are flat and name columns as plain strings. Wide and long data both work. Per-series styling lives in a `seriesConfig` map, where a series also picks its own mark — that is how one chart draws bars and a line together.

## How it is put together

- **ECharts draws only the plot.** The card, title, legend, tooltip, axis titles and the loading, empty and error states are Vue components using semantic tokens, so dark mode and RTL work everywhere. All of that chrome is exported, so an app can build a chart this library does not ship and have it look the same.
- **ECharts comes in through `echarts/core`,** and `registerChartModules` lets an app add any module the library does not register. That registry is what earns the separate entry point, not bundle cost — #870 measured a real consumer build. FunnelChart and NumberCard carry no echarts at all.
- **Three palettes** ship as `--chart-*` tokens, with light and dark values.

## What belongs in the library

`spec/charts-scope.md` states the rule and applies it to a full audit of Insights, the first large consumer. A feature enters because it follows from the library's own conventions, not because an app asked for it. Demand sets the order, never the membership.

That rule is why some obvious-looking props are missing on purpose:

- no legend-position prop — the library decides where the legend sits
- no axis-angle prop — labels fit themselves, tilting only when tilting shows more text
- no quadrant props — a quadrant divider is a reference line
- no pivot table and no map — one is not a plot, the other needs a geography layer the library does not own

## The current charts are going away

`AxisChart`, `DonutChart`, `ECharts`, `FunnelChart`, `NumberChart` and `useAxisChartOptions` still ship from the `frappe-ui` root, exactly as before. **Nothing breaks when this merges.**

A later PR removes them, once apps have moved. This PR adds the new library and changes nothing about the old one. `spec/charts-v1-migration.md` is the migration guide, and `spec/charts-parity.md` lists what did not carry over.

## Docs and tests

`yarn docs:dev` → a dedicated Charts section: an overview, a page per chart with its API reference, a states page and a dashboard example. Foundations → Chart Colors covers the palettes.

518 unit tests over the option builders, 118 component tests in the browser.

## Screenshots

| Light | Dark |
|--------|--------|
| CleanShot 2026-08-02 at 14 49 08@2x | CleanShot 2026-08-02 at 14 50 00@2x |

Coverage: 70.27% (+3.52% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-890/


netchampfaris:
@nextchamp-saqib Add cypress/vitest tests?

netchampfaris:
@greptile review

netchampfaris:
@greptile-apps review

netchampfaris:
[#885](https://github.com/frappe/frappe-ui/issues/885) just resolved on the [1.0.0 map](https://github.com/frappe/frappe-ui/issues/864), and four of its calls land on this PR.

**1. This ships in `1.0.0`, and the old `Charts` family is deleted rather than kept alongside it** ([#942](https://github.com/frappe/frappe-ui/issues/942)). "The current charts stay untouched" holds for the life of this PR, but not past the tag — under map rule 2 an export frozen until `2.0.0` has to be reviewed to bar, and the old family is nowhere near it (no tests, no dark mode at all — `ECharts.vue:28` hardcodes `init(chartDiv, 'light')` — and a raw `EChartsOption` prop that would freeze echarts 5's option object as our public API).

**2. The stated reason for the subpath is wrong, but the subpath is right.** The description says the `echarts/core` split "keeps echarts out of the main bundle". [#870](https://github.com/frappe/frappe-ui/issues/870) measured a real consumer build and found echarts is already dropped from an `import { Button }` — P15 limb 1 does not fire for anything at root. `frappe-ui/charts` stays, standing on limb 2 instead: `registerChartModules` is a genuine extensible registry. Correction handed to [#935](https://github.com/frappe/frappe-ui/issues/935). Nothing to change here beyond the wording.

**3. Combo charts are now blocking** ([#941](https://github.com/frappe/frappe-ui/issues/941)). `spec/charts-parity.md` rates this medium-high effort / high importance; #885 upgraded it to a tag blocker for a reason the audit did not have. crm and helpdesk are **dashboard builders** — chart config is created by their users and stored in the database, so a mixed bar+line chart is reachable in those products today and the app cannot code its way around a shape the library cannot draw. `AxisChartConfig.series` is `(Bar | Line | Area)[]`; the v2 components are homogeneous.

**4. The raw `ECharts` wrapper is not needed.** The parity doc rates it "low effort, high importance" as the escape hatch. Measured against the only consumer that uses it: **all 3 helpdesk sites draw plain category/value bar charts** (`AvgTimeMetrics.vue:338`, `RecentFeedback.vue:546,611`, plus `ChartCardBase.vue` which only forwards a config prop). `BarChart` covers them; `useChart` + `registerChartModules` cover the exotic case. Suggest dropping it from the parity list rather than building it.

Two smaller notes: `charts-style.css` joins the `editor-style.css` / `list-style.css` set that [#903](https://github.com/frappe/frappe-ui/issues/903) is trying to delete, and this entry surface gets its own at-bar sweep at [#944](https://github.com/frappe/frappe-ui/issues/944) — which is where the docs-config gap (`src/charts` is in neither `sourceRoots` nor `colocatedRoots`) and the silent-break migration entries from the parity list are tracked.

PR #912, opened 2026-08-05
fix(editor): stop coercing neutral ink to a hue, and open suggestion menus from a button

netchampfaris:
Two editor fixes that happen to share a branch.

## Black text was turning green

When the editor meets an inline color it does not know, it finds the closest color in its palette and stores that name instead. The problem is that "closest" always returns something. Plain black body text (`#000000`, `#111827`, the legacy `#1F272E`) came out closest to the green anchor, and white came out closest to purple. Whole documents of older content rendered green.

Two changes:

- `getClosestNamedColor` now rejects a match when the input has almost no colour in it. It compares the input's chroma against the matched anchor's: below a third, the input is neutral ink or paper, not a pale version of that hue. Pale palette shades keep about half the anchor's chroma, neutral inks keep well under a fifth, so the threshold sits in open space.
- `matchLegacyHex` used to return `string | null`, which could not tell "this is not a legacy hex" apart from "this is a legacy hex for a colour we deliberately drop". The second case fell through to distance matching and got a hue guessed for it. It now returns a status of `matched`, `excluded` or `unknown`, and `excluded` stops the lookup. Hex matching is also case-insensitive now, because legacy HTML is not consistent about casing.

## Opening the slash and mention menus from a button

`@tiptap/suggestion` has no imperative open. `active` is worked out only by scanning the text before the caret. So a toolbar button has to type the trigger character into the document for real, and then take responsibility for it.

Two ways that went wrong:

- Dismiss the menu (Escape, click away, type a query with no matches) and the trigger stays in your text. The plugin's exit is a meta-only transaction that never edits the document.
- Pick an item and the command's `deleteRange(range)` takes the trigger and the query, but not the space in front of them. That space is needed for the menu to open at all, since a trigger glued to the end of a word never matches. So "Hello world" plus Bullet list left "Hello world " with a blank nobody typed.

`openSuggestionMenu` stamps a marker on the transaction that inserts the trigger. `autoOpenCleanupPlugin`, registered next to every suggester by `createSuggestionExtension`, removes what is left over when the menu closes. A trigger the user typed carries no marker and is never touched.

On the picked path the padding space is only removed when it has nothing left to keep apart: its text block ends there, or the next character is whitespace anyway. A command that inserts inline content instead, such as a mention chip or an emoji, lands right after that space and there it is doing the job a user-typed space would have done. Removing it would glue the insertion onto the previous word.

Ordering note: the cleanup plugin reads the suggester's own state, so it is registered after the `Suggestion` plugin.

## Tests

`suggestion-open.test.ts` is new, 10 tests: opening in an empty paragraph and after a word, dismissing, the whole query run coming back out, a user-typed trigger being left alone, the caret moving away, a typed space killing the match, the padding space taken back when a block command leaves it stranded, the padding space kept when it still separates two words, and the no-suggestion early return.

`color-palette.test.ts` gains cases for neutral ink resolving to no colour, and for real palette colours and their shades still resolving.

`yarn vitest run src/molecules/editor/extensions/shared/` passes, 42 tests.

Docs preview: https://ui.frappe.io/pr-preview/pr-912/

Coverage: 66.56% (+0.08% vs `main`)

### DIFF
--- a/src/molecules/editor/extensions.ts
+++ b/src/molecules/editor/extensions.ts
@@ -307,6 +307,13 @@ export const Tag = TagComposite
 
 export const Emoji = EmojiExtension
 export { SlashCommands }
+
+/**
+ * Opens a suggester's menu from a toolbar button. Exported because there is no
+ * imperative way to do it otherwise — see `suggestion-open` for why the trigger
+ * char has to be typed into the document, and how it is cleaned up again.
+ */
+export { openSuggestionMenu } from './extensions/shared/suggestion-open'
 export const Toc = TocNodeExtension
 export const ContentPaste = ContentPasteExtension
 export const StyleClipboard = StyleClipboardExtension
--- a/src/molecules/editor/extensions/shared/color-palette.test.ts
+++ b/src/molecules/editor/extensions/shared/color-palette.test.ts
@@ -133,4 +133,41 @@ describe('color style', () => {
     // a red-ish hex close to #dc2626
     expect(extractTextColorFromStyle('color: #dd2525')).toBe('red')
   })
+
+  // Regression: nearest-neighbour in RGB space always returns something, so
+  // legacy body ink used to snap to the closest saturated anchor — every
+  // near-black below resolved to `green`, and white to `purple`.
+  it('treats neutral ink as no color instead of the nearest hue', () => {
+    for (const value of [
+      '#1F272E', // legacy frappe/gameplan body ink
+      '#1f272e', // same, lowercased by the source HTML
+      '#000000',
+      '#111827', // tailwind gray-900
+      '#0f172a', // tailwind slate-900
+      '#171717',
+      '#212121',
+      '#333333',
+      '#ffffff',
+      'rgb(0, 0, 0)',
+      'rgb(31, 39, 46)',
+    ]) {
+      expect(extractTextColorFromStyle(`color: ${value}`)).toBeNull()
+    }
+    expect(extractHighlightColorFromStyle('background-color: #f3f4f6')).toBeNull()
+  })
+
+  it('still resolves real palette colors and their shades', () => {
+    expect(extractTextColorFromStyle('color: #16a34a')).toBe('green')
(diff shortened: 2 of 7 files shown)


netchampfaris:
Thanks — the first one was a real bug and is fixed in 5740adc9c, along with the `padded` nit. Notes on all four:

**1. The never-opened leak.** Confirmed and fixed. It was reachable through the `allow: !isInCode(...)` that `createSuggestionExtension` gives every suggester, so the slash button inside a code block left a stray `/`. Two tests cover it now, one padded and one not; both fail against the previous commit.

I did it slightly differently from the suggestion. Rather than deleting `from` → `from + (padded ? 2 : 1)`, the marker now carries the end of the injected span, so the branch takes back exactly what the opener put in and stays correct if the injected text ever changes shape. It appends to the insert's own transaction instead of dispatching a second one, so the insert and its removal collapse into a single undo step.

You were right that the return value was also wrong. `openSuggestionMenu` now reports whether a menu actually opened, and a `false` return also means the document was left untouched.

**2. `padded: text.length > 1`.** Fixed, now `text !== char`. Agreed on the reasoning — no two-char trigger exists today, but the failure mode is silent and eats a real character.

**3. The legacy `color-utils.ts`.** Out of scope on purpose. That copy sits under `components/TextEditor`, which `src/index.ts` marks deprecated in favour of the `frappe-ui/editor` subpath this PR touches. Giving it the same guard means changing behaviour on a deprecated surface with its own consumers and no coverage, which does not belong in the same change as the fix for the current editor. Worth its own issue if the old editor is still expected to render legacy documents correctly.

**4. A command instead of a package export.** Fair point and I like the shape. Not doing it here: the package is on `1.0.0-beta.31`, so the export is not frozen yet, and there is time to move it to `editor.commands.openSuggestionMenu(name)` before v1 without a breaking change. Happy to follow up with that separately.

On the pale-tint note: agreed that red-200 `#fecaca` falls under the threshold. That is the trade I picked, and I would rather lose a rarely used pale tint than render whole legacy documents in green. If it bites, the threshold is one constant.

PR #914, opened 2026-08-05
refactor(editor)!: open a suggestion menu with a command, not an export

netchampfaris:
Follow-up to #912, which added `openSuggestionMenu` as a package export under `frappe-ui/editor`. It should not be one. The function acts on a single editor and belongs with the other editor verbs, so it is now:

```ts
editor.commands.openSuggestionMenu('slashCommands')
// or, from a toolbar button:
editor.chain().focus().openSuggestionMenu('mentionSuggestion').run()
```

`createSuggestionExtension` registers it, next to the cleanup plugin it depends on. Every suggester registers the same generic command; tiptap keeps the last registration of a name and they are identical, because the command takes the suggester to open as an argument rather than belonging to any one of them.

## The insert had to move onto the caller's transaction

A command runs mid-dispatch, and a nested `editor.chain()` there throws `Applying a mismatched transaction`. So the insert is now `insertSuggestionTrigger(tr, char)`, internal to the module, and the command writes to the `tr` it is handed.

## It now decides before inserting

The command runs the same `isInCode` test the suggester's `allow` uses. Pressing the button inside code returns `false` with the document untouched, instead of typing a char and having the plugin take it back out on the next transaction. No insert-then-remove round trip, and no undo-history noise.

The plugin's never-opened branch stays as the safety net for every other reason a match can fail.

## Breaking

`openSuggestionMenu` is no longer exported from `frappe-ui/editor`. Still beta, and the export shipped in `1.0.0-beta.33` less than an hour ago, so nothing outside this repo should be on it yet.

## Tests

15 tests. The plugin behaviours drive `insertSuggestionTrigger` against a hand-wired suggester; a second block drives the command against a real `createSuggestionExtension`, covering open, unknown suggester, an extension with no suggestion, and the refusal inside code.

`yarn vitest run` and `yarn build` both clean.

Docs preview: https://ui.frappe.io/pr-preview/pr-914/

Coverage: 66.56% (±0.00% vs `main`)

### DIFF
--- a/src/molecules/editor/extensions.ts
+++ b/src/molecules/editor/extensions.ts
@@ -307,13 +307,6 @@ export const Tag = TagComposite
 
 export const Emoji = EmojiExtension
 export { SlashCommands }
-
-/**
- * Opens a suggester's menu from a toolbar button. Exported because there is no
- * imperative way to do it otherwise — see `suggestion-open` for why the trigger
- * char has to be typed into the document, and how it is cleaned up again.
- */
-export { openSuggestionMenu } from './extensions/shared/suggestion-open'
 export const Toc = TocNodeExtension
 export const ContentPaste = ContentPasteExtension
 export const StyleClipboard = StyleClipboardExtension
--- a/src/molecules/editor/extensions/shared/suggestion-helpers.ts
+++ b/src/molecules/editor/extensions/shared/suggestion-helpers.ts
@@ -1,4 +1,5 @@
 import type { Editor, Range } from '@tiptap/core'
+import type { Node as ProseMirrorNode } from '@tiptap/pm/model'
 import type { EditorState } from '@tiptap/pm/state'
 
 /**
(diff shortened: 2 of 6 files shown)


netchampfaris:
Three of four were real and are fixed in 3f9374694. The `test` check was red from the flake below; green now.

**1. No focus.** Right, and the PR body example was wrong too. `commands.focus()` inside the command shares its transaction, as you said, so it goes in before the insert. Test blurs first, calls the command, waits a frame (tiptap defers the DOM focus to `requestAnimationFrame`) and asserts `isFocused`. It needs the editor attached to the document — `isFocused` is `document.activeElement === view.dom`, which a detached element can never satisfy.

**2. Stale doc.** Reproduced exactly: `editor.chain().insertContent(...).openSuggestionMenu(...).run()` threw `RangeError: Position 45 out of range`. Rather than swap one argument at the call site I changed `isInCode` to take the doc instead of the state, so the shape of the call makes the pairing obvious and the `allow` site now passes `state.doc`. Both are the only callers.

**3. What `true` means.** Documented on the command type: `true` means the trigger was inserted, not that a menu is showing, and a suggester declining for its own reasons still returns `true` with the cleanup plugin taking the char back out.

**4. Conditional availability.** Leaving this one. It is how every extension command in the repo already behaves: `setVideo`, `reuploadVideo`, the iframe and attachment commands all declare into the global `Commands` interface while existing only when their extension is loaded. Singling this one out would be inconsistent, so I have documented the condition on the type instead. The `SuggestionExtension.configure()` gap is real but not fixable by registering the command there: that path does not install `autoOpenCleanupPlugin` either, so the command would open a menu with nobody to clean up after it. Worth its own issue about whether that older factory should be folded into `createSuggestionExtension`.

Both new behaviours are covered and both tests fail against the previous commit. 231 tests and `yarn build` green.

Issue #692, opened 2026-05-16
[Umbrella] v3 data fetching follow-through and Gameplan migration


netchampfaris:
Post-v1 follow-through on v3 data fetching: full Gameplan migration off v1
resources, ergonomic additions surfaced during migration, Frappe v15 backport,
and the eventual v2 deprecation decision.

v3 itself ships in v1.0 via PR #610 with a frozen public import surface.
This umbrella tracks the work that intentionally did **not** block 1.0.

## In scope

Gameplan migration
- `frontend/src` no longer uses:
  - `resources: {}`, `$resources`, `$getResource`
  - `createResource`, `createListResource`, `createDocumentResource`
- v1 resource usage removed from active app codepaths
- Migration findings folded back into v3 API improvements

v3 ergonomics (additive within frozen surface contract)
- `.submit`, `.cancel`, and similar parity gaps surfaced during migration
- DocType Meta composable (#260)
- Cancel ongoing fetch requests in handles (#26 — re-scope for v3, not legacy resources)
- Additional cache adapter implementations if needed

Backend / framework
- Frappe v15 `/api/v2/*` backport — until this lands, v3 stays v16+ only
- Session and user utilities (#441)
- First-class socket.io utilities
- Broader internal migration of `frappe-ui/frappe/*` to v3

v2 deprecation decision
- Re-evaluate after Gameplan migration completes and v15 backport status is known
- If kept supported, document long-term coexistence; if deprecated, define removal target (likely v2.0)

## Acceptance

- Gameplan fully migrated off v1 resources
- v3 ergonomic gaps from migration are closed
- v15 backport status resolved (landed, or explicitly punted)
- v2's fate decided and documented
- Any additions to v3 honor the v1.x frozen-surface contract (no breaking changes; deprecation cycles only)

## References

- v1 plan: `v1-release/plan.md` — workstreams §4 (carve-out) and §5
- v3 spec: `frappe/client/spec/{api,core,vue}.md` (in PR #610)
- v3 frozen surface contract: see `v1-release/plan.md` Data API strategy section

<details>
<summary>Raw</summary>

1. Ability to append/remove/modify a child row in a document

2. Ability to fetch a "publicly accessible" document via data fetching utility.
```ts
 // for some documents that have `public=1`, allow get doc without permission check
doc = InsightsQuery.get(<>)
// same with a few methods that should be callable without auth/perms
doc.execute()
```

3. First-class doc editing API
doc.isDirty(), doc.discard()

```
[...]

PR #921, opened 2026-08-06
fix(data-fetching): stop a cached doc overwriting fresher data in docStore

netchampfaris:
## The problem

`docStore` gives a consumer the same document twice: once from the IndexedDB cache, once from the server. That part is on purpose, it is stale-while-revalidate. What is missing is any ordering between the two.

Both paths `await` IndexedDB and then assign into the same shared `Ref`. So the stale cached copy can land *after* the fresh server copy and overwrite it. `docRef.value` moves backwards in time, and the component shows old data with no way to tell.

The invariant this PR adds: **`docRef.value` never moves backwards in time.** Consumers still get two publishes. They just always arrive in order.

This is not a design choice elsewhere in the library. `useCall` and `useList` keep the cached copy in a separate ref and use a computed to pick between cached and fresh, so cached can never overwrite fresh in a shared slot. `docStore` was the only primitive merging both into one ref by blind assignment.

## What changed

All in `src/data-fetching/docStore.ts`.

**One assignment point.** Every write to a doc ref now goes through a private `publish()`, which bumps a per-key revision counter. `loadDoc` records the revision before it awaits, and after the cache read returns it only publishes if the revision has not moved. If something newer published in the meantime, the cached copy is dropped.

**`setDoc` publishes before it writes.** It used to `await idbStore.set(...)` first, so readers sat on stale data for a whole IndexedDB round trip, and a failed write hid the new value from them entirely. `setDocs` already assigned before writing, so the two paths disagreed. They agree now.

**Concurrent loads of one key are deduped.** `getDoc` fired `loadDoc` without marking it in flight, so several reads of the same key could run at once and each was free to assign. Worse, `loadDoc` deletes `lastFetched` early, which makes the key look *more* stale to the next reader in the same tick. There is now one in-flight promise per key, cleared when it settles.

**`loadDoc` logs instead of re-throwing.** Nothing awaits it, `getDoc` fires it and returns the ref, so the throw surfaced as an unhandled rejection.

**The store keeps documents raw.** On the cached path `loadDoc` stored the *transformed* doc, while `setDoc` stored raw and `useDoc` transforms on read. A cached doc was therefore transformed twice, so for a non-idempotent `transform` the two publishes carried different values. `getDoc` no longer takes a `transform` argument; callers apply it when they read, which `useDoc` already did.

## Behaviour changes for existing consumers

1. A cached publish that would have clobbered fresh data no longer fires. Anything relying on that was already broken.
2. `setDoc` updates the ref even if the IndexedDB write fails. This matches `setDocs`.
3. On the cached path the ref now holds the raw doc, not the transformed one. Visible to anyone reading `docStore` directly instead of through `useDoc`.
4. `getDoc(doctype, name, transform, options)` is now `getDoc(doctype, name, options)`. The only call site in this repo is `useDoc`, which is updated.

## Tests

Four new tests in `src/data-fetching/docStore.test.ts`. They run under `@vitest-environment node`, where `IDBStore` falls back to an in-memory map, so a controllable deferred on `idbStore.get` is enough to hold a read open.

- a slow cached read never overwrites a newer publish
- concurrent reads of one stale key hit IndexedDB once
- `setDoc` publishes before the IndexedDB write settles
- the store publishes the cached doc raw

All four fail on `main` and pass here. The first one fails with `expected bio: "fresh", received bio: "stale"`, which is the bug itself.

Full suite: 240 tests across 39 files, all passing.

Related to #692.

Docs preview: https://ui.frappe.io/pr-preview/pr-921/

Coverage: 66.64% (+0.05% vs `main`)

### DIFF
--- a/src/data-fetching/docStore.test.ts
+++ b/src/data-fetching/docStore.test.ts
@@ -2,15 +2,35 @@
  * @vitest-environment node
  */
 import { computed, watchSyncEffect } from 'vue'
-import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { docStore } from './docStore'
 import { idbStore } from './idbStore'
 
 const DOCTYPE = 'User'
-const identity = <T>(d: T): T => d
 
 function idbKey(name: string) {
   return `doc:${DOCTYPE}/${name}`
+}
+
+/** A promise whose settlement the test controls, to hold an IDB call open. */
+function defer<T>() {
+  let resolve!: (value: T) => void
+  const promise = new Promise<T>((r) => {
+    resolve = r
+  })
+  return { promise, resolve }
+}
+
+/** Let every pending microtask chain inside docStore run to completion. */
+function flush() {
+  return new Promise((r) => setTimeout(r, 0))
+}
+
+/** Push an entry past the cache timeout, as a long-lived useDoc would be. */
+function makeStale(name: string) {
+  ;(
+    docStore as unknown as { lastFetched: Map<string, number> }
+  ).lastFetched.set(`${DOCTYPE}/${name}`, Date.now() - 10 * 60 * 1000)
 }
 
 /**
@@ -20,7 +40,7 @@ function idbKey(name: string) {
  * this computed the instant setDoc assigns the ref — the trigger for the crash.
  */
 function subscribeLikeUseDoc(name: string) {
-  const doc = computed(() => docStore.getDoc(DOCTYPE, name, identity).value)
+  const doc = computed(() => docStore.getDoc(DOCTYPE, name).value)
   const seen: unknown[] = []
   const stop = watchSyncEffect(() => {
     seen.push(doc.value)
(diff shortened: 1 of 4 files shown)


netchampfaris:
Thanks — the first one is a real regression I introduced, and the reasoning is exactly right. I had considered this case and waved it off with "cleanup deletes the IDB row, so the pending read finds nothing", which is wrong: the `get` was issued before the delete, so it still answers with the row.

Both scenarios now have a failing-first test in `77a775f8a`.

**Deleted doc coming back.** Fixed in three places, because the slot check alone was not enough — after a delete, a *new* `getDoc` opens a fresh slot that the late read would happily write into.

- The revision counter is now store-wide and only ever increments. The per-key counter was the root cause: it was not monotonic, so a stale snapshot of `0` matched a fresh slot's `0`. A guard against values moving backwards has to hold for the guard's own counter.
- `cleanup()` bumps the key's revision instead of deleting it. A delete is an event in-flight reads have to notice.
- `readFromCache` bails when the key has left `docs`, as suggested.

The snapshot also dropped its `?? 0`, so "never published" stays `undefined` and is distinguishable from any real revision.

**staleOnError dedup.** Documented rather than split. Both callers share the ref regardless, and the only difference is the cached copy being kept or dropped one cycle earlier than the second caller asked for. Splitting the read to honour that costs more than it buys — happy to reconsider if you disagree.

**The lastFetched cast.** Fixed, uses `makeStale()`.

Suite is 242 passing.

Issue #917, opened 2026-08-05
CI check: fail when the generated API tables are stale


netchampfaris:
## Question

Nothing to decide — this is work that must happen before the sweeps lean on it.

`docs/scripts/propsgen.ts` generates the props, slots and emits tables from the source into each component's `<Name>.api.md`, and those files are committed. But `yarn docs:gen` runs in **no** workflow. So the committed tables can drift from the code and nobody finds out.

That matters now because [at-bar checklist](https://github.com/frappe/frappe-ui/blob/main/spec/at-bar.md) item 4 deliberately does *not* ask sweepers to hand-verify generated output — it assumes CI keeps the tables honest. Without this job, either the tables drift through the freeze, or fourteen sweeps each waste time re-checking a script's work.

**The work:** add a step to `.github/workflows/tests.yml` (or its own workflow) that runs `yarn docs:gen` and fails if `git diff --exit-code` is non-empty, with a message telling the contributor to run it and commit the result.

**Do first:** run `yarn docs:gen` on a clean tree and commit whatever it produces. The tables may already be stale, and the job should land green rather than red.

**Not in scope:** the 194 of 953 documented members that have an empty description. Those are a source-code gap, not a staleness gap — the generator is emitting exactly what the JSDoc says, which is nothing. Filling them is item 4's work inside each family's sweep.

netchampfaris:
## at-bar item 6 — `PHILOSOPHY.md` audit

Shipped in #945. Scope note first, because it decides most verdicts: this ticket
adds **no public export**. It changes a CI workflow, a docs generator under
`docs/scripts/`, the generated `.api.md` files, and one JSDoc comment inside
`SidebarSection.vue`. `PHILOSOPHY.md` governs the API a consumer imports, so
most principles have nothing to reach. Recording all fifteen anyway, per the
checklist.

| # | Verdict | Note |
| --- | --- | --- |
| P1 | pass | The one name added is a workflow job, `API Tables`. It names the thing being checked, not the mechanics of checking it. |
| P2 | N-A | No state, no component surface. |
| P3 | N-A | No props. |
| P4 | N-A | No colors. |
| P5 | N-A | Not an input control. |
| P6 | pass | No slots added. The regeneration **documents** a slot that was already there and missing from the table — `Sidebar`'s `#sidebar-item`, which is in the shared vocabulary. |
[...]

PR #945, opened 2026-08-06
ci: check the generated API tables against the source

netchampfaris:
Part of #864. Does #917.

## What changed

`docs/scripts/propsgen.ts` writes each component's props, slots and emits into a
`<Name>.api.md` file. Those files are committed to the repo, but nothing ever ran
`yarn docs:gen` in CI. So they could drift out of date with the code and nobody would spot it.

This adds a CI job that regenerates every table and compares it with the committed one.

## The check compares meaning, not bytes

`yarn docs:check` parses both files into rows keyed by name and reports what differs. It fails
when a row is added, removed or renamed, or when its type, default or description changed. It
does not compare row order, or the order of members inside a union.

That is deliberate. vue-component-meta returns props, emits and union members in the order
TypeScript's checker resolved them, and that order shifts with how much of the program is
loaded. A byte diff would fail on a PR that adds an unrelated component folder, because the
folder shifts the order of a table it never touched.

The failure names the file, the component, the row and the field, and prints both values:

```
src/components/Tooltip/Tooltip.api.md
  Tooltip prop `disabled` is missing from the committed table
      description: "Disables the tooltip entirely."
      required: "false"
      type: "boolean"
      default: "false"
  Tooltip prop `hoverDelay` has a different default
      committed: "1.5"
      now:       "0.5"
  TooltipBubble prop `side` has a different type
      committed: "\"left\" | \"right\" | \"bottom\""
      now:       "\"top\" | \"right\" | \"bottom\" | \"left\""
```

Each line is also emitted as a GitHub annotation on the file.

The story list in `<Component>.md` is still compared exactly. It comes off the filesystem, so
it does not move between runs.

## They had already drifted

Regenerating rewrote 59 files and created 3 that had never been generated at all. `Sidebar`
was missing its `#sidebar-item` slot. Several defaults were wrong or missing. `DataImport`,
`Filter` and `HelpCenter` had no table at all.

That regeneration is a separate commit, so the new job starts off passing.

## Generator fixes the check needed first

Anything wrong in the output would have been frozen in by the new check, so these went in
alongside it.

- The output disagreed with Prettier — trailing spaces and extra blank lines that
  `lint-staged` then strips. The next commit touching any generated file would have failed
  the check over formatting instead of a real change.
- A prop declared with an explicit `= undefined` printed `default: 'undefined'`, which the
  table rendered as a literal "undefined". The field is dropped now.
- vue-component-meta reports no description for an emit at all, so every event fell back to
  a generic sentence built from its name — `update:collapsed` came out as "Fired when the
  collapsed changes." Descriptions now come from the emit's JSDoc. Eleven components had a
  written sentence that is now used.
- `SidebarSection`'s `collapsed` description was in a `//` comment, which the generator
  cannot read. Moved to JSDoc.
- `TimePicker`'s `format` prose said "Default: `HH:mm`" but the table showed no default,
  because there is no static one. The prose now says what it falls back to.

## The tables read in source order

vue-component-meta lists props and union members in the order TypeScript's checker resolved
them. String literal types are interned per program, so `"top" | "right" | "bottom" | "left"`
prints as `"left" | "right" | "bottom" | "top"` once another component has claimed those
literals first. The order tracked how much of the program was loaded, not the source.

Every entry reports the file and offset it was declared at, so `docs/scripts/source-order.ts`
reads the order off the AST instead. It resolves an inline union, a type alias, an indexed
access (`TooltipContentProps['side']`), an index into a `const` array or tuple
(`(typeof SIDE_OPTIONS)[number]`), and a runtime prop's `PropType<…>` cast. A printed member
the declaration does not mention means the checker widened the type, so the printed order
stands.

23 union types moved back, across 16 files:

| | before | after | declared as |
|---|---|---|---|
| `Tooltip.placement` | `"left" \| "right" \| "bottom" \| "top"` | `"top" \| "right" \| "bottom" \| "left"` | reka-ui `SIDE_OPTIONS` |
| `Select.size` | `"md" \| "sm" \| "lg" \| "xl"` | `"sm" \| "md" \| "lg" \| "xl"` | `'sm' \| 'md' \| 'lg' \| 'xl'` |
| `Rail.variant` | `"ghost" \| "tile"` | `"tile" \| "ghost"` | `'tile' \| 'ghost'` |
| `Accordion.type` | `"multiple" \| "single"` | `"single" \| "multiple"` | `'single' \| 'multiple'` |
| `Badge.theme` | `"blue" \| "red" \| "green" \| "gray" \| …` | `"gray" \| "blue" \| "green" \| "amber" \| …` | `'gray' \| 'blue' \| 'green' \| 'amber' \| …` |
| `FormControl.type` | `"select" \| "textarea" \| TextInputTypes \| …` | `TextInputTypes \| "textarea" \| "select" \| …` | `TextInputTypes \| 'textarea' \| 'select' \| …` |

The tables carry 125 prop types with a top-level union. The other 102 were already in
source order.

Rows were not all in declared order either. `Button` listed
`theme, variant, type, label, loading, size, …` where `buttonProps` declares
`theme, size, variant, label, icon, …`. `ListView`, `ListRowItem` and `Radio` were out too.
Rows are now sorted by the offset they are declared at, and a file is kept together in the
order it first appears — so `DatePicker` lists `CommonDatePickerProps`, then its own props,
then the ones it inherits from `InputLabelingProps`.

Some unions have no declaration to read: `RouteLocationRaw` and `Exclude<InputVariant,
'ghost'>` are not written as unions, and `defineModel<boolean | null>()` reports no
declaration at all. Those keep the checker's order, which already matches the source in every
case here.

A per-folder run was the alternative — that path already produced source order. It costs 170s
against 3.8s for one full run, and it still gets `Button` wrong, so it was not worth it.

## Worth knowing

`yarn docs:gen Badge` still does not always produce the same file as a full run. Props and
union members no longer move, but emits can: an event declared by `defineModel()` is declared
in Vue's own types, and where that lands relative to the ones from `defineEmits<T>()` is the
checker's call. Nine of 64 files differ that way; before source order, 25 did. The check
accepts either, so a partial run will not fail CI, but a full run before committing is still
worth it, and a partial run prints a reminder.

A union nested inside another type is still checker-ordered — only a prop's own type is
reordered. One row is visibly wrong because of it: `list`'s sort slot prints
`{ direction: "desc" | "asc" | null }` where the source says `'asc' | 'desc'`. Reordering a
union nested in a printed object type means re-printing the type, which is a much larger
change than reading the declaration, so it is left as is.

## What this breaks

Nothing. No export was removed or renamed, so there is no changelog or migration entry.

`SidebarSection` now declares `update:collapsed` in `defineEmits` as well as `defineModel`,
so the event has somewhere to carry a description. Vue merges the two; the emitted event is
unchanged.

## Checks

`yarn test` — 376 pass, including 29 for the check and 14 for source order.

`yarn type-check` — 168 errors, the same 168 already on `main`. None of them are in files this
PR touches.

`yarn docs:gen` twice in a row leaves a clean tree. `yarn docs:check` passes on it. A full run
takes 3.58–4.03s, against 3.61–3.96s before source order — no measurable cost.

Source order, checked against the source. Of the 125 prop types with a top-level union, 101
are declared as an inline union; every one of the 101 lists its members in the order the
committed table shows. The other 24 have no inline union to read — 20 go through an alias, a
`PropType` cast or an indexed access, and 4 report no declaration. `Button`, `ListView`,
`ListRowItem`, `Radio` and `DatePicker` match their prop declarations row for row.

Pass case, on real output: a per-folder run of all 64 folders wrote 9 files that differ byte
for byte from the committed full-run output. `yarn docs:check` passed on all 9.

Fail case, on real output: five mutations to `Tooltip.api.md` — a prop added, removed,
renamed, retyped, and its default and description changed. Each was reported, with the
component, the row, the field and both values. Also checked: a table that has never been
generated, a table that is empty while the component has props, and a stale story list.

Coverage: 66.98% (+0.31% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-945/


netchampfaris:
## Review — NEEDS-WORK

The job itself is well built and the regeneration is real work. Three claims the PR
rests on do not survive checking, and one of them is the reason the nondeterminism
was left in place.

---

### 1. "The full run's order is the source's order" is false, and this PR ships a docs regression that the new check freezes in

This is the premise the PR uses to decline sorting the unions. It does not hold.

`Badge`'s `theme`, the exact example the PR names:

- source (`src/components/Badge/types.ts:7`) — `'gray' | 'blue' | 'green' | 'amber' | 'red' | 'violet' | 'orange'`
- committed here, from a **full** run — `"blue" | "red" | "green" | "gray" | "amber" | "violet" | "orange"`
- `yarn docs:gen Badge`, the **partial** run the PR warns against — `"gray" | "blue" | "green" | "amber" | "red" | "violet" | "orange"`

The partial run reproduces the source exactly. The full run scrambles it. I ran both
on a clean checkout of `1009c7b2b` to confirm.

And the regeneration replaced correct orders with scrambled ones in the published
tables. Each of these was right before this PR:

| table | source order | committed now |
| --- | --- | --- |
| `Progress` / `Select` `size` | `sm \| md \| lg \| xl` | `md \| sm \| lg \| xl` |
| `Tooltip` `side` | `top \| right \| bottom \| left` | `left \| right \| bottom \| top` |
| `Accordion` `type` | `single \| multiple` | `multiple \| single` |
| `Tree` connectors | `connectors \| lines \| none` | `none \| connectors \| lines` |
| `Rail` `variant` | `tile \| ghost` | `ghost \| tile` |

The cosmetic half is the smaller half. The order depends on how much of the program
the checker loaded, and folder traversal is what decides that. So adding or renaming
**any** component folder can reorder unions in unrelated tables — and the new job
will fail a PR that touched nothing near them. That is the false positive that
teaches contributors to distrust a generated-file gate, and it is the failure mode
this ticket exists to prevent.

Sort the union members (or otherwise normalize before writing). The reason given for
not doing it is the opposite of what the generator actually produces.

### 2. A drift finding that is not in the diff

> `Popover` passes `open` to its slot. The table did not list it.

It does not. `Popover.api.md` changes by exactly one added blank line —
`git show ddbb97ccc:src/components/Popover/Popover.api.md` and the file at HEAD have
byte-identical `slotsData` blocks. No slot prop was added.

Across all 53 modified tables the only members actually added are `sidebar-item`
(`Sidebar`), a whole `PageHeaderTitle` block (`title`, `default`), and `placeholder`
+ `item-create` (`frappe/Link`). The `Sidebar` claim is real. The `Popover` one is not.

This matters past the PR body: the item 6 audit on #917 uses it as the evidence for a
**pass** on P7. That verdict is standing on nothing.

### 3. The bug it says it caught, one line lower, unfixed

`SidebarSection`'s `collapsed` prop description moved into JSDoc — good. The
`update:collapsed` emit directly below had the same problem and was not touched, so
regeneration swapped a written sentence for the generator's fallback:

```
- description: 'Fired when the section is collapsed or expanded.'
+ description: 'Fired when the collapsed changes.'
```

That string comes from the `Fired when the ${eventKey} changes.` template at
`docs/scripts/propsgen.ts:204`. Ungrammatical text, now published, and the new check
locks it in. Same class of defect the PR flags to sweepers as a shape to watch for.

### 4. The counts do not match the diff

"rewrote 55 files and created 3" — the diff is **53** modified `.api.md` plus 3 added
= 56 generated files. 59 is the total including `tests.yml`, `propsgen.ts` and the
`.vue`. The #917 comment then says "I diffed every member name across all 59
regenerated tables"; there are 56.

The conclusion is right — I re-ran that member diff across every table and no prop,
slot or emit disappears anywhere — so at-bar item 9 N/A on both halves is correct.
But the number offered as the evidence is not the number that was checked.

### 5. Do not merge on local verification alone

`gh run list --branch fix/ci-api-tables-freshness` returns zero runs. The workflow
file has not been parsed by GitHub once. The queue claim is true and not this PR's
fault — `Tests` runs on other branches are sitting queued 3h56m+ — but given finding
1, "it is deterministic" was established entirely on one macOS machine. The `API
Tables` job needs one green run before this lands.

---

### Checked and confirmed good

- `yarn test` — re-ran it, 39 files / 242 tests, all pass.
- `yarn type-check` — 171 errors on the branch and **171 on `main`**. I ran both. The
  "already failing, not caused here" claim holds.
- A full `yarn docs:gen` on a clean tree of this branch leaves no diff, so the check
  is green here.
- Branch is off `main` at `ddbb97ccc`, 5 commits, nothing foreign. It is 4 commits
  behind current `main` (docs only).
- Break classification is right, and for a stronger reason than the one given: the
  only source file touched is `SidebarSection.vue` and that hunk is comment-only.
  No consumer behavior can change, so there is nothing to classify. Not the expensive
  mistake.
- The job covers both gaps the triage comment named — the generator runs as its own
  step so a per-folder throw (`process.exitCode = 1`) fails the build, and
  `git add --intent-to-add` catches new tables.
- The PR description reads plainly and quickly. The problem with it is accuracy, not
  clarity.

### Minor

- `default: 'undefined'` is now printed for `Sidebar`'s `active` and `isActive`. That
  renders a literal "undefined" in the defaults column.
- `TimePicker`'s `format` **lost** `default: '"HH:mm"'`, while its description still
  says "Default: `HH:mm`". The removal is defensible — there is no prop default, the
  value is computed and becomes `h:mm A` under `use12Hour` — but the table and the
  prose next to it now disagree, and the PR lists this under "defaults were wrong or
  absent" as though it were a fix.

Issue #933, opened 2026-08-06
Take the v2 composables to bar: tests and the P1-P15 audit


netchampfaris:
## Question

Take the five public v2 composables to bar: `useList`, `useDoc`, `useCall`, `useDoctype`, `useNewDoc`.

Decided on [#886](https://github.com/frappe/frappe-ui/issues/886) — v2 is the recommended data layer at `1.0.0`, so it carries the full weight of [`spec/at-bar.md`](../blob/main/spec/at-bar.md). Blocked by [#931](https://github.com/frappe/frappe-ui/issues/931): auditing and testing a shape that is about to change is wasted work.

These are non-component exports, so the shorter list applies — items 1, 2 and 5 are N/A, item 4 is discharged by [#932](https://github.com/frappe/frappe-ui/issues/932), and item 7 is trivially satisfied (nothing here is marked `@deprecated`).

**Item 3 — tests.** Current state: `useList.test.ts` 255 lines, `useCall.test.ts` 253, `useDoc.test.ts` 178, `useDoctype.test.ts` 84, and **`useNewDoc` has no test at all**. Adapt the five behaviors to a composable: returns sane state before the first fetch, the reactive round-trip (a change to a reactive input refetches and the result shows up), loading and error states behave, and every documented option actually does what the page says.

**Item 6 — the P1–P15 audit.** A verdict list on this ticket. Worth a hard look at the naming and return shapes specifically, because #886 froze these five names at the root export until `2.0.0` and [#870](https://github.com/frappe/frappe-ui/issues/870) ruled out a `frappe-ui/data` subpath, so a rename later is not available.

**Item 9** — a changelog line, plus a migration before/after for anything whose break is silent.

Note the adoption context when weighing findings: 19 call sites across 13 files, all in gameplan. The cost of getting a name wrong here is paid by future users, not by a migration.

netchampfaris:
## The adoption figure in this ticket's body is wrong

The last paragraph reads:

> Note the adoption context when weighing findings: 19 call sites across 13 files, all in
> gameplan. The cost of getting a name wrong here is paid by future users, not by a
> migration.

**Both sentences are false.** Re-counted today against freshly cloned trees with imports
resolved: **132 sites across 119 files in five apps**. Full table and method on
[#886](https://github.com/frappe/frappe-ui/issues/886#issuecomment-5221464397).

| app | commit | `useDoc` | `useList` | `useDoctype` | `useNewDoc` | `useCall` | sites | files |
[...]

Issue #931, opened 2026-08-06
Fix the shared-useCall concurrency bug in useDoctype and useList


netchampfaris:
## Question

Fix the shared-`useCall` concurrency bug across the v2 composables, before their shape freezes at the tag.

`useDoctype`'s `delete`, `setValue`, `runDocMethod` and `runMethod` (`src/data-fetching/useDoctype/useDoctype.ts:47,76,126`) and `useList`'s `setValue` and `delete` (`src/data-fetching/useList/useList.ts:179,199`) each hold a **single** `useCall` instance whose `url` is a `ref` mutated at submit time. Two concurrent submits therefore share one `loading` / `error` / `data` and race on the URL — the second submit overwrites the first's target before the first resolves, so both resolve against the last URL set.

Decided on [#886](https://github.com/frappe/frappe-ui/issues/886): v2 is the recommended data layer at `1.0.0`, so this has to be right before the freeze. The fix likely changes what these methods return, and that shape is frozen until `2.0.0` — which is why this gates the v2 docs and the v2 at-bar work rather than running alongside them.

Cover the fix with tests. `useDoctype.test.ts` is 84 lines today and does not exercise concurrent submits.

netchampfaris:
Done in #946.

## What shipped

Every `submit()` on `useDoctype`'s `insert` / `delete` / `setValue` /
`runDocMethod` / `runMethod` and on `useList`'s `insert` / `setValue` /
`delete` now builds its own URL and sends its own request. The eight methods
went from carrying the whole `useCall` surface to this, which is what freezes
at the tag:

```ts
{
  submit(params): Promise<Response | null>  // this submit's own response
  data                                      // last submit that succeeded
  error                                     // last submit that settled
  loading                                   // true while any submit is running
  isLoading(target)                         // true while this one target is running
}
```

Removed: `params`, `promise`, `url`, `reset`, `abort`, `execute`, `fetch`,
`reload`, `isFetching`, `isFinished`, `canAbort`, `aborted`. Each described the
one shared request that no longer exists.

`params` is the removal that matters. Its only real use is
`delete.loading && delete.params.name === row.name` — the "which row is
deleting" check — and that check is already wrong today: start two deletes and
both rows spin. `delete.isLoading(name)` and `setValue.isLoading(name)` replace
it and stay right when submits overlap. `runDocMethod.isLoading(name, method)`
and `runMethod.isLoading(method)` keep their signatures; they used to compare
[...]

Issue #932, opened 2026-08-06
v2 export surface and docs pages, and split the docs sidebar


netchampfaris:
## Question

Settle the v2 export surface and give it the docs it has never had, then split the docs sidebar so "recommended" is visible to a reader rather than only to `plan.md`.

Decided on [#886](https://github.com/frappe/frappe-ui/issues/886). Blocked by [#931](https://github.com/frappe/frappe-ui/issues/931) — the concurrency fix changes what several of these return, and documenting the current shape would mean rewriting the pages.

**Export changes:**

- **Un-export `useFrappeFetch`.** It is purely the internal building block `useCall`, `useDoc` and `useList` sit on (`useCall.ts:3`, `useDoc.ts:10`, `useList.ts:15`), with zero downstream call sites across all six apps counted. Freezing it until `2.0.0` buys nothing.
- **Export `FrappeResponseError`.** v2 throws it (`src/data-fetching/useFrappeFetch.ts`) but does not export it, so consumers cannot type a catch block. Close the gap while the surface is still movable.

**Docs.** `plan.md` claims v2 is "documented in main docs" and v1 is "hidden from standard docs". Both halves are inverted: `docs/.vitepress/config.ts:90` holds one Data Fetching section with exactly three pages — Resource, List Resource, Document Resource, all v1 — and the v2 composables appear nowhere on the site.

Split the sidebar in two, with no "legacy" or "deprecated" wording in either, since #886 settled that neither is true:

- **Data Fetching** — `useList`, `useDoc`, `useCall`, `useDoctype`, `useNewDoc`, with a line naming it the recommended layer for new code.
- **Resources** — the three existing v1 pages, with a line stating it is fully supported through `1.x`.

Leaves a clean slot for v3 when it lands in a `1.x` minor.

Per at-bar item 4, every prop, option and return member on the new pages needs a non-empty description, and the prose and examples have to be true.

netchampfaris:
Resolved via #972 (merged). v2 data-fetching docs complete with a split sidebar; staleOnError docs corrected to match canUseCachedFallback() (FrappeResponseError carve-out); three silent bugs fixed test-first along the way (useNewDoc reactivity, initialData on useCall and useList).

Issue #886, opened 2026-07-30
Data API export posture for 1.0.0


netchampfaris:
## Question

Settle the **data API** export posture for `1.0.0`.

Three generations ship simultaneously. Confirm or revise what `v1-release/plan.md` says, and make sure the code matches whatever is decided.

**v1 resources** (`src/resources/` — `createResource`, `createListResource`, `createDocumentResource`, `resourcesPlugin`). Plan says: still exported, supported as a migration path, deprecated for new usage, hidden from standard docs, warns in dev. But `deprecated-removals.md` notes they are **not actually marked `@deprecated` in code**, so ADR-0008 does not reach them. Decide whether that is deliberate — marking them pulls them into the pre-tag deletion sweep, which is a much bigger change.

**v2 composables** (`src/data-fetching/` — `useList`, `useDoc`, `useCall`, `useDoctype`, `useNewDoc`). Plan says: exported, **not** deprecated for 1.0, recommended for Frappe v15 and existing codebases, no dev warnings, in the main docs. Confirm that survives.

**v3 client** (`frappe-ui/frappe/vue` — `createClient`, `createDefaultCacheAdapter`, `defineDoctype`, doc/list/count/newDoc handles). Ships in 1.0 via PR #610, recommended for Frappe v16+.

Open items:

- PR #610 has been open since March. Establish its actual state — merged, stale, or still pending self-review against `frappe/client/spec/*.md`.
- v3's import path is `frappe-ui/frappe/vue`, but `frappe-ui/frappe` is being removed. **These collide.** v3 needs a home that survives; coordinate with the `frappe-ui/frappe` and subpath tickets.
- Document the v3 frozen surface contract: what is stable for 1.x versus what may evolve via deprecation cycle only.
- The one-time dev-mode notice on `createClient` ("v3 is recommended for Frappe v16+").
- Known bug to fold in: `useDoctype`'s `runMethod` is a single shared `useCall`, so concurrent calls cross-resolve.

netchampfaris:
**Fact for the v3 import-path question.** Checked freshly fetched `develop` for crm, helpdesk, builder, insights, and gameplan: **no app imports `frappe-ui/frappe/vue`** — zero consumers of the v3 path today.

That means the collision with #867 (removing `frappe-ui/frappe`) costs nothing to fix right now. Moving v3 to a path that survives the removal is currently a free change; it stops being free the moment `1.0.0` freezes the import path. Worth doing before the tag rather than after.

[...]

PR #946, opened 2026-08-06
fix(data)!: give every useDoctype and useList submit its own request

netchampfaris:
Part of #864. Does #931.

## What was wrong

The write methods on `useDoctype` (`insert`, `delete`, `setValue`, `runDocMethod`, `runMethod`)
and on `useList` (`insert`, `setValue`, `delete`) each held **one** request, shared by every call
to `submit()`.

So when two submits overlapped, the first was cancelled but still returned a result — either the
other submit's answer, or nothing at all.

`delete` was worse. Its success handler read the shared settings, so two deletes at once removed
the wrong row from the stored data.

Gameplan has already run into this. `frontend/src/data/unreadCount.ts` carries a hand-written
queue and a comment that ends "the real fix belongs in frappe-ui's useCall".

## What changed

Every submit now builds its own address and sends its own request, so two submits can no longer
interfere with each other. Tests came first — seven of them failed before the fix went in.

Here is what you get back now, and this is the shape that freezes at 1.0.0:

```ts
{
  submit(params): Promise<Response | null>
  data
  error
  loading
  isLoading(target)
}
```

These are gone: `params`, `promise`, `url`, `reset`, `abort`, `execute`, `fetch`, `reload`,
`isFetching`, `isFinished`, `canAbort` and `aborted`. Each of them described the single shared
request, which no longer exists.

`isLoading(name)` replaces the old `delete.params.name === row.name` trick. That trick is already
wrong today — start two deletes and both rows show a spinner.

```vue
<!-- before, from gameplan's TaskList.vue -->
<div :class="{ 'pointer-events-none': tasks.delete.loading && tasks.delete.params.name === d.name }">

<!-- after -->
<div :class="{ 'pointer-events-none': tasks.delete.isLoading(d.name) }">
```

`submit()` has one outcome channel: it resolves with the response, or rejects with the error.
A failed `validate` already rejected; a failed request used to resolve with `null`. Both reject
now, so `null` stops meaning "it failed" and a server that answers with `null` resolves with
`null`. Callers that checked `if (!result)` need a `try` / `catch` or a `.catch()`.

```js
// before
const doc = await todos.insert.submit({ title: 'Buy milk' })
if (!doc) return showError(todos.insert.error)

// after
try {
  const doc = await todos.insert.submit({ title: 'Buy milk' })
} catch (e) {
  showError(e)
}
```

Two more quiet ones, both because `data` and `error` no longer come from a shared request:

- `data` is not reset to `null` when a submit fails. It keeps the last successful response.
- `error` is not cleared when a submit starts. It stands until the newest submit settles.

```js
await todos.setValue.submit({ name: 'TODO-1', status: 'Done' })
await todos.setValue.submit({ name: 'TODO-2', status: 'Done' }) // fails

todos.setValue.data // before: null — after: still the TODO-1 response
todos.setValue.error // the failure

// so test `error`, not `data`
if (!todos.setValue.data) {
} // before: "the last save failed" — after: never true once one save has worked
```

Also fixed: `useList`'s `insert` and `delete` ignored the `baseUrl` you passed to `useList`.
`setValue` already used it, so all three now agree. `useDoctype` was never affected.

```js
const todos = useList({ doctype: 'ToDo', baseUrl: 'https://other.site' })
await todos.insert.submit({ title: 'Buy milk' })
// before: POST /api/v2/document/ToDo on the current origin
// after:  POST https://other.site/api/v2/document/ToDo
```

## Who this breaks

Six Frappe apps import `useDoctype` or `useList`: gameplan, central, suite, writer,
frappe_calendar and frappe-ui-starter. Three of them are hit.

**Gameplan** — the widest, and undercounted before. Three `delete.params.name` expressions in
`Settings/InvitePeople.vue` and `TaskList.vue`, plus about twenty sites across a dozen files
that assume a failed `submit()` resolves: floating submits with no `.catch()`, and guards like
`if (comments.insert.error || !comment?.name) return` that can no longer run.
frappe/gameplan#538.

**frappe_calendar** — `EventEdit.vue:516` reads `events.insert.data` after the submit,
`EventDetail.vue:37-39` deletes with a `.then()` and no `.catch()`, and the three wrappers in
`data/calendarEvent.ts` flip from resolve-null to reject. frappe/frappe_calendar#9.

**suite** — one site: a bare `signatures.delete.submit()` in
`mail/components/Settings/SignatureSettings.vue:110`. frappe/suite#541.

Not hit: **central** uses `useCall`, which this PR does not touch. **writer** and
**frappe-ui-starter** only read from `useList`; their writes go through `useDoc`, deferred
to #933.

**Grepping for the member names is not enough.** v1's `createListResource` exposes
`delete.params.name`, `insert.data` and `submit()` under exactly the same names, and this PR
does not touch it. crm, studio, builder, press, lms, helpdesk, hrms, wiki and draw all match
on those strings; none of them is affected. Check what built the resource before counting a
hit.

**Nothing will warn you.** The build passes and the app looks fine until someone deletes a row.
So this gets a changelog line and a before-and-after in the migration guide.

## Not in this PR

`useDoc`'s `setValue` and `delete` have a smaller version of the same problem — two saves of the
same document still cancel each other. Noted on #931 and left for #933.

## Checks

`yarn test` — 317 pass across 43 files.

`yarn type-check` — 171 errors, the same 171 already on `main`. None in `src/data-fetching`.

Docs preview: https://ui.frappe.io/pr-preview/pr-946/

Coverage: 67.26% (+0.59% vs `main`)


netchampfaris:
## Review — independent check of #931 / PR #946

The fix is real and the core of it is right: every submit now builds its own URL and
its own request, the tests genuinely fail before the fix, and the base and call-site
counts check out. Five things below, worst first.

---

### 1. `error` is never cleared on success, so a stale error outlives a later successful submit

`src/data-fetching/useAction.ts` — `error.value` is written in exactly two places
(the validate branch, and the failure branch after the await) and cleared in exactly
one (`error.value = null` at the top of `submit`). The success path sets `data` and
returns without touching `error`:

```ts
    let response = await call.submit(body ? body(params) : params)
    let callError = (call.error ?? null) as Error | null
    if (callError) {
      error.value = callError
      return null
    }
    data.value = response as TResponse   // error left alone
    return response as TResponse
```

Repro, with two submits overlapping — which is the case this PR exists to make correct:

1. `delete.submit({ name: 'a' })` — fails fast (permission error).
2. `delete.submit({ name: 'b' })` — started at the same time, slower, succeeds.

`b` settles last and succeeded, but `delete.error` still holds `a`'s error and stays
that way until the *next* submit starts. A `<ErrorMessage :message="tasks.delete.error" />`
sits there showing an error for a delete that worked.

This contradicts all three places the new shape is written down:

- `useAction.ts` docblock: "`error` and `data` come from the submit that settled last"
- `v1-release/changelog.md`: "`error` is cleared when a submit starts and set from the
  submit that settles last"
- the PR body: "`error` — error of the last submit that settled"

The mirror problem is the clear-on-start: submit B wipes submit A's error while A is
still in flight. Between the two, aggregate `error` has no definition that survives
overlap — and `error` is one of the five members freezing until `2.0.0`. Worth deciding
now: settle-order tracking, or say plainly in the docs that `error` is last-failure and
never last-settled.

### 2. `insert` has no `isLoading`, so the "frozen shape" is wrong for two of the eight methods

PR body, changelog and the #931 comment all state the frozen shape as:

```ts
{ submit(params), data, error, loading, isLoading(target) }
```

and the #931 comment says "the eight methods went ... to this". They didn't.
`useInsert` in `useDoctype.ts` and `insertAction`/`insert` in `useList.ts` pass no `key`
and expose no `isLoading` — six methods have it, two don't. Since an added member
freezes exactly like a renamed one (ADR-0012, restated on #864), "add it later" is not
available after the tag. Either give `insert` a key and `isLoading`, or write down why
insert is the one action with no per-target answer — and fix the three documents that
currently claim otherwise.

### 3. at-bar item 4 is marked N/A on a misreading

The #931 comment marks items "1, 2, 4, 5, 8: N/A. Non-component exports". Item 4's
N/A column does not say that. It says non-component exports "document on the utilities,
composables or subpath page instead — no colocated page, no playground". The docs
requirement itself still applies.

`useDoctype` and `useList` are documented nowhere on the site today (#886 established
this), and this PR removes 12 members and adds a new one without a doc page. That's
item 4 **not done and deferred to #932**, which is the honest split exit — not N/A.
The checklist's own words: an item is either done or written down as N/A, never skipped.

### 4. No CI evidence on the PR

`gh pr checks 946` shows one check: Greptile. `gh run list --branch fix/usecall-concurrency`
is empty, so `tests.yml` never ran on this branch even though it triggers on
`pull_request`. The green claim carried no automated proof.

I re-ran it in the worktree and it holds:

- `yarn test` — 39 files, 254 tests, all pass. `useDoctype.test.ts` 11, `useList.test.ts` 12.
- `yarn type-check` — 171 errors on the branch, 171 on `origin/main`, `grep 'data-fetching'`
  matches none of them.

So the numbers are honest. But someone should find out why Actions isn't firing on this
branch before it merges.

### 5. `data` on failure changed, and isn't recorded

Old behavior: `useCall.data` read through to VueUse's `data`, which VueUse resets to
`initialData || null` on any non-ok response (`@vueuse/core` `useFetch`, the
`if (!fetchResponse.ok)` branch). So a failed submit left `data` at `null`.

New behavior: `data.value` is only ever written on success, so after a failure it still
holds the previous successful response. Anything rendering `insert.data` after a failed
insert now shows the last good one instead of nothing. Silent, and in neither the
changelog nor `migration.md`.

---

### What I checked and found clean

- **Base.** `git merge-base origin/main origin/fix/usecall-concurrency` = `3d429bb48`,
  which is `origin/main` HEAD. Exactly 4 commits, all on topic. No unrelated work rode in.
- **Ticket coverage.** All six locations #931 named are fixed (`useDoctype` `delete`,
  `setValue`, `runDocMethod`, `runMethod`; `useList` `setValue`, `delete`), plus `insert`
  on both. Nothing from the ticket was silently skipped.
- **Rule 5.** Re-counted against a freshly fetched `origin/develop`, not a local checkout.
  gameplan gives exactly three hits and they are the three claimed —
  `Settings/InvitePeople.vue:74`, `TaskList.vue:34`, `TaskList.vue:41`, all the
  `delete.params.name` idiom. helpdesk `origin/develop` is 0. Count is not stale.
- **`useAction` is genuinely internal** — absent from `src/data-fetching/index.ts` and
  `src/index.ts`, so the item 0 "internal" call stands.
- **Item 9.** Silent is the right call — the removed members are read from `.vue`
  templates, which no build step type-checks, so the app builds and renders and then
  throws a `TypeError` on the first delete. Both the changelog line and the
  `migration.md` before/after are there. Correct, and correct on the expensive side.
- **PR description.** Plain English, leads with the problem, readable in well under
  thirty seconds. No complaints.

netchampfaris:
Two decisions recorded on [#933](https://github.com/frappe/frappe-ui/issues/933) that
bound this PR. Neither asks for a change here.

**1. The read side keeps its names.**
[Verdict](https://github.com/frappe/frappe-ui/issues/933#issuecomment-5221475472):
`reload` / `fetch` / `execute`, `loading` / `isFetching`, plus `isFinished`, `canAbort`,
`aborted`, `abort` and `url` all ship unchanged on `useCall`, `useDoc` and `useList`.

This PR's P1 verdict — that `isFetching` / `canAbort` / `aborted` name transport rather
than state — is sound for the write side, where those members had to go anyway once the
shared request was deleted. It is deliberately not extended to the read side. There those
members still describe a real request, so removing them is a pure naming change: a tidier
surface bought with a migration in every app that uses one.

**2. "Not in this PR" stands.** The `useDoc` / `useNewDoc` half of the bug stays out of
#946 and ships as its own PR under #933, before the tag —
[scope](https://github.com/frappe/frappe-ui/issues/933#issuecomment-5221481354). The
`TaskDetail.vue` dropped-write in your #931 note is confirmed against gameplan `a25bf81`.
`useAction` is the tool, and it stays unexported.

Also relevant to the migration sections of this PR body: the v2 downstream census on #886
was badly low. It is **132 import sites across 119 files in five apps**, not 19 / 13 in
gameplan — and **central** (40 `useCall` sites, zero v1) was missing from the app list
entirely. Full table:
[#886](https://github.com/frappe/frappe-ui/issues/886#issuecomment-5221464397). This PR's
own six-app list is the more accurate one; central uses `useCall` only, so its "not hit"
call still holds.

Issue #944, opened 2026-08-06
Take charts v2 to bar


netchampfaris:
## Question

Take charts v2 (`frappe-ui/charts`) to the [at-bar checklist](https://github.com/frappe/frappe-ui/blob/main/spec/at-bar.md).

[#885](https://github.com/frappe/frappe-ui/issues/885) decided v2 ships in `1.0.0`. Map rule 2 then applies to it like any other family: exported at `1.0.0` means frozen until `2.0.0`, so it gets the same review every other export gets. The sweep tickets #871–#884 were charted before v2 existed, so no ticket covers it.

Members: `AreaChart`, `BarChart`, `DonutChart`, `FunnelChart`, `HeatmapChart`, `LineChart`, `NumberCard`, `useChart`, `registerChartModules`, `currentColorScheme`, `paletteColors`, `resolveChartTheme`, `useChartTheme`, `formatAxisValue`, `formatDate`, `formatLabel`, `formatPercent`, `formatValue`, plus the exported types.

Blocked on combo support, which changes the frozen prop shape.

v2 starts well ahead of the families swept so far — it already has unit tests, docs pages, exported types and theme-reactive palettes. The real work is items 0, 6 and 9:

- **Item 0** — an explicit keep-or-un-export call per member. The format helpers (`formatDate`, `formatValue`, `formatLabel`, `formatPercent`, `formatAxisValue`) and the theme members are the ones to look hard at: several are building blocks that may have no consumer-facing reason to be public. `registerChartModules` is a bare re-export of echarts' `use` and is worth naming deliberately, since it is also the reason the subpath is justified at all (see below).
- **Item 6** — the P1–P15 verdict list. `registerChartModules = use` is the load-bearing one: it is what earns `frappe-ui/charts` its separate entry point under P15's registry limb, now that #870 killed the bundle-cost limb. See the note handed to [#935](https://github.com/frappe/frappe-ui/issues/935).
- **Item 9** — `spec/charts-parity.md` lists several changes that are **silent** for anyone porting from v1 and so need a `migration.md` before/after, not just a changelog line: `connectNulls` default flipped `true` → `false`; `formatDate`'s argument order swapped, which breaks positional callers with no compiler error; `formatValue` shorten precision default `0` → `1`; `NumberCard` value formatting changed from always-compact to full-precision; and v1's forced `min-w-[300px] min-h-[300px]` floor removed, so a chart in an unsized parent renders at zero height.
[...]

Issue #941, opened 2026-08-06
Add combo series support to charts v2


netchampfaris:
## Question

Add combo series support to charts v2 (`frappe-ui/charts`) so one chart can mix bar, line and area series.

Decided on [#885](https://github.com/frappe/frappe-ui/issues/885): the old `Charts` family is deleted at the tag and v2 replaces it. Combo is the one capability v1 has that v2 does not, and it is the only parity gap on `spec/charts-parity.md` that a real consumer can reach.

**Why it blocks the removal.** `AxisChartConfig.series` is typed `(BarSeriesConfig | LineSeriesConfig | AreaSeriesConfig)[]` (`src/components/Charts/types.ts:35`) — a single `AxisChart` mixes series types. v2's `BarChart`, `LineChart` and `AreaChart` are homogeneous. Two of the three chart consumers (crm, helpdesk) are **dashboard builders**: the chart config is created by their users and stored in the database, not written by their developers. So a combo chart is reachable today by anyone using those apps, and dropping the shape is not something the consuming app can decide its way out of.

Combo plus `y2Axis` is also the canonical "value vs rate" pairing, which is the reason a second y-axis exists at all — v2 already shipped `y2Axis`.

Scope:

- mixed series types in one chart, however the v2 prop shape expresses it (`seriesConfig` already carries per-series styling)
- works with the existing `y2Axis` / `series.axis` support
- tests and a docs example
- `spec/charts.md` updated, and the combo row cleared from `spec/charts-parity.md`

This changes the v2 prop shape, so it lands before v2 is frozen at `1.0.0` — hence it blocks both the removal and the v2 sweep.

netchampfaris:
Done, in #950 (draft, against `charts-v2`).

**The shape.** `seriesConfig.<key>.type` — `'bar' | 'line' | 'area'` — with the
host chart's own shape as the default. The ticket's guess that `type` belongs in
`seriesConfig` rather than in a new prop held up: it keeps a saved combo chart
one object spread into one tag, which is the whole reason this blocks the tag
(crm and helpdesk store the config, they don't write it). All three axis charts
can host a combo, so converting a saved v1 `AxisChart` picks the tag from the
majority series and recasts the rest. Works with `y2` — the recast series takes
the second axis and its format, which is the value-vs-rate chart y2 exists for.

**Two things the ticket and the audit did not have.**

1. **v1 never drew a horizontal combo — it threw.** `axisChartOptions.ts:18`
[...]

PR #950, opened 2026-08-06
feat(charts): combo series — one chart, mixed shapes

netchampfaris:
One chart can now draw bars, lines and areas together. This was the last thing the old charts
could do that charts v2 could not.

Targets `charts-v2` (#890), not `main`. Does #941. Decided on #885.

## What changed

A series is drawn as its chart's own shape, unless its `seriesConfig` entry names a `type`:

```vue
<BarChart
  :data="rows"
  x="month"
  :y="['revenue', 'expenses']"
  y2="margin"
  :series-config="{ margin: { type: 'line', label: 'Gross margin' } }"
  :y2-axis="{ title: 'Margin', format: (v) => `${v}%` }"
/>
```

`type` goes in `seriesConfig` rather than in a new prop or a fourth component. That way a saved
chart stays one object spread into one tag, which is how crm and helpdesk actually store them.
The three axis charts differ only in the default shape they hand their series, so any of them
can host the same mix.

Underneath, the three option builders became one. Anything the mix decides is decided across
every series, not just the visible ones, so clicking the legend never reshuffles the plot: any
bar means a shaded hover pointer, the label gutter is as wide as the widest series needs,
`stacked` adds each shape into its own stack, and a line always draws on top of the bars it is
read against.

It works with the second axis, and on a horizontal bar chart — where the old charts threw an
error, so this is better than what came before rather than equal to it.

Tests: 20 option tests and 4 component tests. Docs: a revenue-against-margin story on the
BarChart page. `spec/charts.md` has the API and the four decisions behind it.

## What this breaks

**Nothing that ships today.** `charts-v2` has never been released.

- **Nothing fails to build.** `seriesConfig` was widened, so configs written before this still
  type-check.
- **One quiet change inside the branch:** echarts stack names now carry the shape as a prefix
  (`stack` becomes `bar:stack`). You can only reach this through per-series `echartOptions`.
- **A cost, not a break:** all three axis charts now load both the bar and line echarts modules,
  because `seriesConfig` decides the shape while the page runs. A page using only `LineChart`
  now also carries the bar module. `charts.md` said the opposite and is corrected.

## What I did not do

- **The legend still draws a dot for every series.** The old charts used echarts' own legend,
  which draws a different icon per type. Making a mixed chart the one exception seemed worse than
  keeping it consistent. Flagged for #944.
- **`connectNulls` and `fillOpacity` stay chart-wide.** A line inside a bar chart can only reach
  them through per-series `echartOptions`. Not worth freezing a wider set of props at 1.0.0 for a
  rare case.
- **No changelog entry.** The whole v2 surface gets one entry when #890 merges.

## Checks

`yarn test` — 544 pass across 52 files.

`yarn type-check` — no new errors.

`yarn test:cypress` on `src/charts/*.cy.ts` — 67 pass.

`yarn docs:build` runs out of memory on this machine both before and after the change, so the new
docs page was checked through the dev server instead.

Coverage: 68.05% (+1.39% vs `main`)

### DIFF
--- a/spec/charts-parity.md
+++ b/spec/charts-parity.md
@@ -2,17 +2,18 @@
 
 Audited 2026-08-01, comparing `src/components/Charts` (v1) against `src/charts` (v2).
 Dual axis (`y2Axis` + `series.axis`) has landed since the audit, as has `HeatmapChart`,
-which v1 never had.
+which v1 never had. **Combo charts have landed too** ([#941](https://github.com/frappe/frappe-ui/issues/941)):
+`seriesConfig.<key>.type` draws a series as a bar, line or area whichever axis chart
+hosts it, so v1's mixed `AxisChart.series` has a shape to convert to. See
+[`charts.md`](./charts.md) for the API and the decisions behind it.
 
 ## Missing components
 
 1. **Raw `ECharts` wrapper** — v1's documented escape hatch for chart types we don't
    ship (gauge, sankey, …). v2 only has the `useChart` composable. Low effort, high
-   importance.
-2. **Combo charts** — v1's `AxisChart` mixed `bar`/`line`/`area` series in one chart;
-   v2 components are homogeneous. Combo + y2 is the canonical "sales vs growth rate"
-   chart and the main reason y2 exists. Medium-high effort, high importance.
-3. `useAxisChartOptions` export — replaced by per-chart builders with a different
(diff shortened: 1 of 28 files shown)


netchampfaris:
Closing: @nextchamp-saqib will land combo series in his own PR. The shape decision on [#941](https://github.com/frappe/frappe-ui/issues/941) (`seriesConfig.<key>.type`, host shape as default) and this diff are there to draw from.

PR #952, opened 2026-08-07
chore(deps): bump @vueuse/core to ^14.1.0

### LINKED ISSUE #895: Duplicate @vueuse/core: frappe-ui pins ^10, reka-ui needs ^14
`package.json` declares `@vueuse/core` `^10.4.1` while the pinned reka-ui requires `^14.1.0`. The ranges cannot unify, so package managers install two copies and bundlers ship both (~35 kB gzip each, plus a second `@vueuse/shared`).

Suggestion: bump frappe-ui's own dependency to `^14` so one copy satisfies both. The v10→v14 changes that touch frappe-ui's usage are small.

adityapradhan10:
## Summary
- Bump `@vueuse/core` from `^10.4.1` to `^14.1.0` so it matches `reka-ui`'s requirement
- Removes the duplicate VueUse install (`10.x` + `14.x` plus a second `@vueuse/shared`) that package managers could not unify
- No source API changes; frappe-ui already peers `vue >= 3.5.0`, which VueUse 14 requires

Fixes #895

## Test plan
- [x] Confirm `yarn.lock` resolves a single `@vueuse/core@14.1.0` (no `^10.4.1` entry)
- [x] `yarn type-check` — no VueUse-related errors (existing unrelated errors remain)
- [x] `yarn test` — data-fetching / `createFetch` suites pass; one flaky editor timeout passes on re-run

Made with [Cursor](https://cursor.com)

Coverage: 66.66% (±0.00% vs `main`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
     "@tiptap/starter-kit": "^3.26.0",
     "@tiptap/suggestion": "^3.26.0",
     "@tiptap/vue-3": "^3.26.0",
-    "@vueuse/core": "^10.4.1",
+    "@vueuse/core": "^14.1.0",
     "codemirror": "^6.0.1",
     "dayjs": "^1.11.13",
     "dompurify": "^3.4.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1625,11 +1625,6 @@
   resolved "https://registry.yarnpkg.com/@types/unist/-/unist-3.0.3.tgz#acaab0f919ce69cce629c2d4ed2eb4adc1b6c20c"
   integrity sha512-ko/gIFJRv177XgZsZcBwnqJN5x/Gien8qNOn0D5bQU/zAzVf9Zt3BlcUiLqhV9y4ARk0GbT3tnUiPNgnTXzc/Q==
 
-"@types/web-bluetooth@^0.0.20":
-  version "0.0.20"
-  resolved "https://registry.yarnpkg.com/@types/web-bluetooth/-/web-bluetooth-0.0.20.tgz#f066abfcd1cbe66267cdbbf0de010d8a41b41597"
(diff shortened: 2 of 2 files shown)


netchampfaris:
Thanks for opening this, and sorry it sat for a while.

You were right about the bump, and about the reason. Since you filed it, #1045 turned up the concrete failure this prevents: because we pinned `^10.4.1` while reka-ui needs `^14.1.0`, a consumer who adds `resolve.dedupe: ['@vueuse/core']` collapses reka-ui onto our 10.x copy, where `useResizeObserver` tests `Array.isArray()` against a still-wrapped `ComputedRef` and hands the whole array to `observe()`. It throws on every Tabs instance, app-wide. That made this a release blocker for 1.0.0 (#1059) rather than a cleanup.

What was missing was the audit to justify a four-major bump: a check of every `@vueuse/core` call site in the repo, and proof that the nested copies actually disappear. I have done that in #1066 and it confirms your position. Nothing we use was renamed or removed, `vue-tsc` reports an error set byte-identical to main, 1079 unit tests pass against a real 14.1.0, and a scratch install shows 4 copies of `@vueuse/core` collapsing to 1.

Your `package.json` and `yarn.lock` edits are identical to what landed there, and I have credited you in the PR body and in the issue. #1066 supersedes this one only because it carries the audit alongside the same diff, so I would like to close this in its favour, but I will leave that to a maintainer rather than doing it here.

Two small notes on the description, for the record: the version that resolves is 14.1.0, not the current registry latest, because the lockfile already carried that entry for reka-ui. And the duplicate was wider than it looked. `vitepress` and `@vueuse/integrations` each carried a nested copy too, so it was four copies rather than two.

Thanks again for spotting it early.

adityapradhan10:
Thanks for the thorough follow-up, and for taking the time to audit this properly.

Happy to close this in favour of #1066. The `package.json` / `yarn.lock` change is the same, and that PR carries the call-site audit and the proof that the nested copies collapse. Glad the bump also turned out to be the right fix for the Tabs / `useResizeObserver` failure in #1045.

Thanks for the credit on the issue and in #1066.

PR #956, opened 2026-08-07
refactor(overlays)!: bring Popover, Tooltip, HoverCard and BottomSheet to the v1 bar

### LINKED ISSUE #872: Sweep: overlays
## Question

Bring the **overlay** family to bar for `1.0.0`.

Members: `Popover`, `Tooltip`, `HoverCard`, `BottomSheet`.

Apply the at-bar checklist to each. Family-specific concerns:

- `Popover` is called out in `v1-release/plan.md` as "the last floating outlier": refactor to the v1 floating vocabulary — `v-model:open`, `side` / `align` / `offset` (deprecating `placement`), `data-slot` styling hooks instead of `popoverClass` (`P10`), canonical slot names, and a11y. `spec/popover.md` is the contract.
- `spec/hover-card.md` is the `HoverCard` contract.
- `BottomSheet` was never classified — keep-or-unexport call needed, plus whether it is the mobile expression of an existing overlay rather than its own component.
- Cross-overlay vocabulary: `CONTEXT.md` says all overlays bind visibility via `v-model:open`, and `dismissible` is named for cross-overlay reuse but ships only on Dialog in v1. Confirm or revisit.
- Shared floating utilities: is there one consolidated implementation, or per-component drift?
- `usePopoverMotion` is a shared composable — check whether it is public surface.

`Dialog` is deliberately a separate ticket; keep the vocabulary aligned with it.

netchampfaris:
Brings the overlay family — `Popover`, `Tooltip`, `HoverCard`, `BottomSheet` —
to the v1 bar. Closes #872.

Breaking, and deliberate: `Popover`'s v0 layer and `NestedPopover` are gone
rather than deprecated. ADR-0008 says nothing marked `@deprecated` ships in
1.0.0.

## Popover

- Removed the v0 back-compat layer: `placement`, `show`, `update:show`,
  `popoverClass`, `#target`, `#body`, and the rest. `side` / `align` / `offset`
  / `v-model:open` / `#trigger` / `#default` are the API.
- `Popover` no longer inherits attributes onto a wrapper. Put them on your own
  element inside `#trigger`.

## NestedPopover

- Deleted. It never nested anything and was the last `@headlessui/vue` +
  `@popperjs/core` popover in the library. `@popperjs/core` leaves
  `dependencies` with it.
- `ListFilter` moves onto `Popover`.

## Tooltip

- `placement` → `side`, matching `Popover` and `HoverCard`.
- `arrowClass` is removed (P10). Style the arrow through `[data-slot="arrow"]`.
  It was documented as the arrow's fill but every call site used it to nudge
  the bubble's position — `offset` does that directly now.
- New `offset` prop. New `[data-slot]` hooks on content, bubble and arrow.
- `TooltipBubble` is no longer exported. It is internal to `Tooltip` and
  `Button`.

## HoverCard

- Added `open()` and `close()` on the template ref, matching `Popover`.
- Typed the trigger slot props.

## BottomSheet

- Kept and documented — 8 call sites across central, gameplan and suite. It
  gains docs, three stories and a contract test file; it had none.
- Fixed a focus bug. The sheet opts out of autofocusing its first field so it
  does not pop the keyboard on a phone. That left focus on the trigger behind
  the overlay, and reka only pulls escaping focus back to the last element
  focused *inside* the sheet — so `Tab` walked the page behind an open modal.
  The sheet now takes focus itself. The keyboard still stays down.

## Bugs found while rewiring

- `CalendarWeekDayEvent` passed `placement="center"`. `center` is an alignment,
  not a side, and the old resolver handed it to reka as a side.
- `MonthPicker`'s panel rendered with no surface. Its `popover-class` had
  already become a no-op.
- `ListFilter`'s filter row could sit past the viewport edge on a narrow
  screen. `NestedPopover` rendered inline so the page could scroll to it;
  `Popover` portals and fixes its panel, so the row was unreachable. The panel
  now caps at the available width and scrolls.

## Migration

Before/after for every break is in the migration guide, under
[Popover / HoverCard / Tooltip](docs/content/docs/migration.md).

## Tests

`Popover` 18, `Tooltip` 7, `HoverCard` 8, `BottomSheet` 9, `ListFilter` 5,
`MonthPicker` 5 — all passing. `vitest`: 376 passing. Type-check is unchanged
at the 168-error baseline `main` already carries.

Coverage: 67.33% (-0.11% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-956/


netchampfaris:
Closing the loop on the review:

- **Popover `dismissible` only covered outside-click.** Real. `:dismissible="false"` still closed on Escape. Fixed by wiring `@escape-key-down` as well.
- **Tooltip `#body`.** Right call. Measured across all 10 downstream apps: `#content` had 0 users, and 6 of the 7 `#body` users hand-copied the bubble's classes to rebuild the surface `#body` had just removed. `#body` is now `#content` (renders inside the bubble) plus `bare` for content that brings its own surface. `#body` was never in the slot vocabulary — PHILOSOPHY.md P6 now says so.
- **BottomSheet focus.** A test caught a genuine gap: preventing `open-auto-focus` left the focus trap with no anchor, so Tab walked the page behind an open modal. The sheet now focuses itself, which keeps the original intent (no mobile keyboard on open) and restores the trap.

Downstream call sites are listed per app on the 10 `Upgrading to frappe-ui 1.0.0` issues.

Issue #871, opened 2026-07-30
Sweep: selection and menus


netchampfaris:
## Question

Bring the **selection and menu** family to bar for `1.0.0`.

Members: `Select`, `Combobox`, `MultiSelect`, `Dropdown`, `ItemListRow`, `ContextMenu`, `Menu`.

Apply the at-bar checklist to each. Family-specific concerns:

- `spec/selection.md`, `spec/dropdown.md`, and `spec/item-list-row.md` are the existing contracts — verify the code matches them, and fix whichever is wrong.
- `ItemListRow` is the shared row primitive; each higher-level component owns its own listbox shell. Confirm that boundary actually holds in the code and has not eroded.
- Vocabulary alignment across the family: `v-model`, `v-model:open`, `@update:query`, and the trigger/item slot names. `CONTEXT.md` says query stays internal by default.
- `Menu` and `ContextMenu` were never classified in `v1-release/plan.md`. Keep-or-unexport call needed for both, plus how they relate to `Dropdown` — three menu-ish components is likely one too many (`P8`).
- `MultiSelect` has an open PR: #809 (creatable support).

## Not in this ticket

The whole `Autocomplete` removal — the decision, the internal `ListFilter.vue` / `SearchComplete.vue` migration, and the deletion — belongs to [#926](https://github.com/frappe/frappe-ui/issues/926), which blocks this. Settled by [#869](https://github.com/frappe/frappe-ui/issues/869): the removal spans this family and `FormControl`, so it lands as one change set rather than split across two sweeps.

This ticket also **loses its "additive fixes from the `Autocomplete` research" bullet** — all four gaps that research found (`bodyClasses`, `maxOptions`, the `items` group key, `#target`) have zero call sites inside or outside the repo, so nothing is added. Per [ADR-0012](../blob/main/spec/adr/0012-imperative-api-sign-off.md), an added prop freezes until `2.0.0` exactly like a renamed one.

netchampfaris:
Context from [Read insights' Autocomplete call sites](https://github.com/frappe/frappe-ui/issues/868) (resolved) — four `Combobox` / `MultiSelect` gaps land in this sweep as additive fixes. Additive, so none of them blocks the freeze, but all four are cheaper to settle while this family is open.

[...]

PR #957, opened 2026-08-08
refactor(Dropdown)!: remove placement, { group, items } and component rows

netchampfaris:
Part of the selection-and-menus sweep (#871) on the [Road to 1.0.0](https://github.com/frappe/frappe-ui/issues/864) map. The pickers were frozen in #858; this PR brings the menu half to bar.

## ADR-0008 removals

| Removed | Replacement |
| --- | --- |
| `placement` prop, `DropdownPlacement` type | `align` |
| `{ group, items }` group entries | `{ group, options }` |
| `component:` option rows, `DropdownComponentOption` / `ContextMenuComponentOption` | `slots: { item: fn }` |
| `DropdownExposed` type | nothing — it described an expose that never existed (ADR-0012) |

The three behavioral removals are **silent breaks** in plain JS (old code runs, renders wrong). Mitigations: the removed keys stay in the types as `never` so TS callers break loudly, a dev-only console warning fires when the old shape reaches the menu at runtime, and the migration guide has before/afters.

## Fixes

- **Disabled state now reaches reka's trigger.** Previously only the generated `Button` was natively disabled; a custom trigger slot with a `disabled` attribute could still be opened by keyboard or synthetic clicks.

## Tests

Five-behavior gaps closed (at-bar item 3): keyboard nav, disabled items + trigger, `#empty`, group labels, `v-model:open` for Dropdown and ContextMenu; `#prefix`/`#suffix`/`data-size` for ItemListRow. 37 menu-family cypress tests green, pickers re-run green (135), vitest 424 green.

## Docs

- `spec/dropdown.md` stopped mandating the deprecated aliases (it predated ADR-0008 and lost), documents `matchTriggerWidth`, and drops the never-implemented `emptyText` prop.
- Migration guide: new [Dropdown and ContextMenu](https://github.com/frappe/frappe-ui/blob/forge/sweep-selection-menus/docs/content/docs/migration.md#dropdown-and-contextmenu) section.
- Changelog entries + deprecation-log rows flipped to **Removed**; also corrected the stale claim that Select's `option` slot-prop alias still ships (it was removed in #858).
- Dropdown, ContextMenu and ItemListRow barrels spell out their exports (P15).

Docs preview: https://ui.frappe.io/pr-preview/pr-957/

Coverage: 67.23% (-0.02% vs `main`)

### DIFF
--- a/docs/components/recipes/TasksDesktop.vue
+++ b/docs/components/recipes/TasksDesktop.vue
@@ -993,12 +993,12 @@ function addComment() {
               <span class="text-sm text-ink-gray-5">
                 {{ visibleTasks.length }} tasks
               </span>
-              <Dropdown :options="groupByDropdownOptions" placement="right">
+              <Dropdown :options="groupByDropdownOptions" align="end">
                 <Button variant="ghost" icon-left="lucide-layers">
                   Group: {{ groupByLabel }}
                 </Button>
               </Dropdown>
-              <Dropdown :options="sortDropdownOptions" placement="right">
+              <Dropdown :options="sortDropdownOptions" align="end">
                 <Button variant="ghost" icon-left="lucide-arrow-up-down">
                   {{ sortLabel }}
                 </Button>
--- a/docs/content/docs/migration.md
+++ b/docs/content/docs/migration.md
@@ -137,6 +137,64 @@ Option values are `string | number` everywhere. `Select` no longer accepts
 | `displayValue` slot prop | `summary` on `#summary`, or `selectedOptions`                       |
 | `toggleOpen` slot prop   | `setOpen(boolean)`                                                  |
 
+### Dropdown and ContextMenu
+
+| Before                                     | After                            |
+| ------------------------------------------ | -------------------------------- |
+| `placement` prop, `DropdownPlacement` type | `align`                          |
+| `{ group, items }`                         | `{ group, options }`             |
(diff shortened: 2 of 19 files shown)


netchampfaris:
Rebased onto `main` (was conflicting after the overlays and toggles sweeps merged). One conflict, in `v1-release/changelog.md`'s deprecation-log table — both sweeps flipped different rows to **Removed**; resolved by taking both sets.

The rebase also surfaced a regression on `main`: #958 removed `Switch.change` and `labelClasses` but missed Menu's internal switch rows, which still bound both — so `ContextMenu`'s switch-item test fails on `main` today (`onClick` never fires; verified on a pristine `main` worktree, and that the test passes at `be5912b10~1`). Fixed here in 36ffcc0cc since Menu is this PR's family: the row's Switch now binds `update:model-value` and drops the dead prop. Menu-family suites green: Dropdown 17, ContextMenu 14, ItemListRow 6.

netchampfaris:
Nits addressed in 96a478e6c: leftover `placement` now DEV-warns (it was the only removal without one), the two menu warnings dedupe per message instead of firing per row per recompute, and `spec/dropdown.md` records the deliberate `matchTriggerWidth` difference from Popover (`width` vs `minWidth`).

Not taken: the pickers' `emptyText` — agreed it's a follow-up, and per ADR-0012 it's frozen with #858's surface, so removing it would be its own decided break, not a nit.

PR #1001, opened 2026-08-09
fix(tokens): emit oklch from sync-tokens

### LINKED ISSUE #986: sync-tokens silently reverts colors.json from oklch to hex
## Question

`yarn sync-tokens` regenerates `tailwind/colors.json` in hex, reverting the manual oklch conversion (fa18b8ade was never folded into the generator). Confirmed: 313 oklch values in colors.json, 0 in generated/. Fold the conversion into the generator so a routine sync can't undo it.

netchampfaris:
Fixes #986.

`yarn sync-tokens` regenerated `colors.json` in hex, reverting the manual oklch conversion (fa18b8ade was never folded into the generator).

- Add a hex → oklch converter to `figma-tokens-to-theme.js` (Björn Ottosson's OKLab reference math).
- Convert every literal color value: primitive ramps, neutral, overlays, 8-digit alpha hex, and literal pass-throughs in semantic styles. Alias references are untouched.
- Verified: regenerated `tailwind/colors.json` is byte-identical to the hand-converted file; `generated/colors.json` now carries 0 hex values.

No API change. Consumer-visible values are unchanged.

Docs preview: https://ui.frappe.io/pr-preview/pr-1001/

Coverage: 59.28% (±0.00% vs `main`)


netchampfaris:
Addressed all three barista nits:

- 63c1b71a8 — `hexToOklch`/`toOklch` are exported, `main()` runs only when invoked as a CLI (same guard as `migrate-tokens-v2.js`), and a vitest locks in the output for `#ffffff` and an 8-digit alpha hex.
- a0ce249fb — the converter throws on hex values that are not 6 or 8 digits, so shorthand fails loudly instead of emitting `oklch(NaN NaN NaN)`.
- 07dc3da63 — deleted `scripts/figma-variables-to-colors.js`; it wrote raw hex into `tailwind/colors.json` and nothing referenced it.

Issue #904, opened 2026-08-04
Move spritePlugin, IconPicker and the sprite Icon to frappe-ui/experimental


netchampfaris:
## Question

`frappe-ui/icons` currently earns its subpath (ADR-0010) only because `spritePlugin` statically imports the full `lucide-static` sprite as a raw string and injects it into the DOM at app startup. Nothing else in the subpath has a dependency. Remove that dependency and the whole subpath folds into root.

`spritePlugin` exists for exactly one consumer: `IconPicker` reads `<symbol>` tags out of the injected sprite DOM (`icons/IconPicker.vue:49`) to build its list of pickable icons. Nothing else needs the sprite — icons render everywhere else through the `lucide-*` Tailwind mask-class convention (`tailwind/lucideIconsPlugin.js`), which needs no DOM injection.

Work:

- `tailwind/lucideIconsPlugin.js` already resolves `lucide-static`'s icon directory at build time. Emit that same name list as a plain export `IconPicker` can import, instead of reading it from injected DOM.
- Rewire `IconPicker` to the generated list; render each option as `<span class="lucide-x">` instead of `<Icon :name>` against the sprite.
- Delete `spritePlugin` and its DOM injection.
- Delete `icons/Icon.vue`. It duplicates the root `Icon` component under the same name with a different mechanism (`<use href="#name">` against the sprite vs. the `lucide-*` mask-class convention) — a live naming bug independent of this cleanup.
- Audit the 8 one-off icon components in `icons/`: `CircleCheckIcon` (used by `Toast`) and `DownSolidIcon` (used by legacy `ListView`) are live; `GreenCheckIcon`, `LightningIcon`, `MaximizeIcon`, `MinimizeIcon`, `StepsIcon` have no internal or downstream callers found on fetched `upstream` refs — confirm and delete the dead ones under ADR-0008.
- Known API gap surfaced while checking downstream usage: crm and helpdesk each independently carry a hand-written comment (`ViewModal.vue`) working around `IconPicker` being lucide-only — their views can store an emoji, which the picker can't select. Worth deciding whether `IconPicker` should support emoji directly (root's `Icon` component already does) as part of this rewire, since two apps have already patched the same gap.

[...]

Issue #887, opened 2026-07-30
Build-time and tooling subpaths


netchampfaris:
## Question

Settle the **build-time and tooling subpaths** for `1.0.0`.

Members: `frappe-ui/icons`, `frappe-ui/tailwind`, `frappe-ui/tailwind/tokens.js`, `frappe-ui/vite`, `frappe-ui/vitepress`, `frappe-ui/experimental`, `frappe-ui/code-editor`, the `*-style.css` entries, and `frappe-ui/tsconfig.base.json`.

These are exported and therefore freeze at the tag, but none has been through a v1 review. They are also a different *kind* of surface from components — mostly build-time rather than runtime — which the subpath rule needs to have accounted for.

Per subpath, decide:

- does it ship in `1.0.0`, and under what name?
- what exactly is its public surface, and is any of it accidental?
- is it documented?

Specific questions:

- **`experimental`** — the name promises instability, but exporting it still freezes it. Either state plainly that it carries no stability guarantee (and that this is a named exception to the freeze), or stop shipping it.
- **`code-editor`** — `plan.md` claims `CodeEditor` is "not exported", which is **stale**: it has had its own subpath for some time. Reconcile the docs, then decide whether it earns a subpath under the new rule or folds into root.
- **`tailwind`** — the plugin drives the `lucide-*` icon class form and the semantic design tokens, so it is load-bearing for the whole library. Confirm its exported surface is deliberate. Also check that the content globs exclude build artifacts.
- **`vite`** — includes the barrel-import rewrite plugin that turns `from 'frappe-ui'` into deep imports. Confirm it is documented, since consumers depend on it for build performance.
- **`vitepress`** — is this library surface at all, or docs-site infrastructure that should not ship?
- **`icons`** — what is in it, and how does it relate to the `Icon` component and the `lucide-*` string convention.

Blocked on the subpath rule.

netchampfaris:
## Resolution

**Build-time entries freeze additive-only** — options, tokens, utilities and compiler options may be *added* in a minor; nothing may be renamed or removed before `2.0.0`. ADR-0010 opened this category ("decided on their own terms") without saying what the terms are; this is them. A strict component-style freeze would end the design token vocabulary on the day we tag and kill the Figma sync; a no-promise carve-out is not something six apps whose CSS and builds depend on the preset would accept.

[...]

PR #1002, opened 2026-08-09
refactor(icons)!: move spritePlugin, IconPicker and sprite Icon to frappe-ui/experimental

netchampfaris:
Part of #864. Does #904 (redirected: keep, don't delete).

Apps still use `spritePlugin`, `IconPicker`, and the sprite `Icon` (crm, insights, suite, press). Deleting them before 1.0.0 breaks those apps. This PR moves them to `frappe-ui/experimental` instead, outside the 1.0.0 freeze, so app authors get time to migrate.

- `frappe-ui/icons` keeps only the named SFC icons.
- `frappe-ui/experimental` now exports `Icon` (sprite), `IconPicker`, `spritePlugin`.
- Migration for apps: change the import path. `lucide-*` classes remain the canonical way to render icons.

Supersedes #947.

Docs preview: https://ui.frappe.io/pr-preview/pr-1002/

Coverage: 66.68% (±0.00% vs `main`)


netchampfaris:
All four points addressed:

- `tailwind/content.js` now globs `experimental/SpriteIcons/**` as an explicit exception (migration parking spot for a supported surface), and the printed list in the Tailwind docs page matches (36f5511b2).
- ADR-0010: the icons-subpath row no longer cites the sprite import. Note: after the move there is no literal name collision left — the row now stands on limb (c)'s forward-looking clause plus the historical `Icon` collision (d4a61c0ed).
- Changelog and migration note now say styles keep working for apps that spread `content` from `frappe-ui/tailwind`, and add the `Icon` alias line (292f7e5e5).

netchampfaris:
Both nits fixed: content.js comment now states the real invariant (c89772a52), and the unwired story file is gone along with the tsconfig exclude it forced (46b8e38f1).

Declining the third: folding `frappe-ui/icons` into root. Maintainer call (Faris, on this PR's review thread): the subpath stays, and the ADR row's wording is approved as-is. See also #887's resolution.

PR #1013, opened 2026-08-09
refactor(tokens): migrate internal rounded-* aliases to numbered radius (#997)

netchampfaris:
Mechanical vocabulary sweep per ADR-0006. Pixel values are identical — no visual change. Unblocks the alias removal in #998.

Refs #997

## Mapping

| Old alias | New token | px | Replaced |
|---|---|---|---|
| `rounded` | `rounded-4` | 8 | 240 (incl. 25 directional: `-t/-r/-b/-l`) |
| `rounded-sm` | `rounded-1` | 4 | 51 (incl. 2 `rounded-tl-sm`) |
| `rounded-md` | `rounded-5` | 10 | 45 (incl. 1 `rounded-r-md`) |
| `rounded-lg` | `rounded-6` | 12 | 58 (incl. 2 `rounded-t-lg`) |
| `rounded-xl` | `rounded-7` | 16 | 11 |
| `rounded-2xl` | `rounded-8` | 20 | 4 (incl. 1 `rounded-t-2xl`) |

Total: 409 replacements across 189 files (`src/`, `docs/`, `icons/`, `experimental/`, `vitepress/`, `spec/`, `skills/`, `tailwind/plugin.js` componentStyles, `App.vue`).

## Deliberate keeps

- `rounded-none` and `rounded-full` — retained per ADR-0006.
- `tailwind/generated/radius.json` alias definitions — #998 removes them.
- `spec/adr/0006-numbered-radius-tokens.md` and the deprecation table in `spec/foundations.md` — they document the aliases.
- `docs/content/docs/migration.md` and `v1-release/changelog.md` — historical migration docs.
- One historical quote in `spec/popover.md` ("legacy `rounded-lg border ...` shell").
- Prose uses of the word "rounded" ("rounded corners", "rounded hover surface") in comments and specs.
- `src/components/ListFilter/` — #999 deletes the directory.

## Tests

- vitest: 54 files, 470 tests passed.
- Cypress component suite: full component suite green — 68 specs, 860 tests passed (one earlier run had 5 flaky failures while a type-check ran concurrently; clean on rerun).
- `yarn type-check`: same pre-existing errors as main (vitepress/ and VNodeChild typing); no new errors.

Docs preview: https://ui.frappe.io/pr-preview/pr-1013/

Coverage: 64.23% (±0.00% vs `main`)

Issue #998, opened 2026-08-09
Remove deprecated radius aliases and text-*-black; teach the codemod radius renames


netchampfaris:
Part of #864. Decided in #993. Blocked by #997 (internal alias sweep).

1. Extend `tailwind/migrate-tokens-v2.js`: rename `rounded`/`rounded-sm/md/lg/xl/2xl` → numbered tokens (deterministic map, ADR-0006). Flag `font-black` for manual attention instead of merging onto a `text-*-black` class.
2. Remove the alias keys from the theme build: `RADIUS_NAME_BY_PX` aliases in `tailwind/figma-tokens-to-theme.js`, regenerate `tailwind/generated/radius.json`. Keep `none` and `full`.
3. Drop `black` from `WEIGHT_VARIANTS` in `tailwind/plugin.js` and the `tailwind.config.js` safelist; remove the Black option from the docs typography page weight switcher.
4. `text-10xl`–`12xl` stay (decided: design-system-defined).
5. Migration guide: radius section (silent break — unmigrated aliases emit no style), `font-black` note. Breaking PR title per rule 11.

netchampfaris:
**Resolved in [#1014](https://github.com/frappe/frappe-ui/pull/1014)** (`refactor!`; based on `v1/997-radius-token-sweep`, retargets to main after #1013 merges).

All five spec items done: codemod radius renames for all directional/logical sides (idempotent, string/`@apply`-scoped so prose "rounded" is untouched; 8 new test cases), `RADIUS_NAME_BY_PX` down to `none`+`full` with `radius.json` regenerated, black weights out of `WEIGHT_VARIANTS`/safelist/docs switcher, `text-10xl`–`12xl` untouched, migration-guide radius section with the silent-break warning plus the `font-black` flag note.

**Dogfood validation** (codemod run on pre-sweep main, diffed against #1013's manual sweep): **0 codemod bugs, 0 manual-sweep misses**; the only divergence is 8 meta-files a consumer tree doesn't have (ADR/spec/migration docs, ListFilter).

Two surprises raised: (1) radius renames broke the codemod's "run once" assumption — an app that already migrated colors/typography would double-shift sizes on a re-run — solved with a documented `--radius-only` mode; flag if you'd rather sequence it differently. (2) #1013 kept ListFilter's two aliases for #1011's deletion, but #1011 is unmerged — #1014 migrates those two lines so no merge order ships silently-missing radii.

[...]

Issue #993, opened 2026-08-08
Token vocabulary: three now-or-never decisions at the tag


netchampfaris:
## Question

P15's additive-only freeze starts at the tag, so these are now-or-never (#940, PR #976):

1. Remove the deprecated radius aliases per ADR-0006? Blast radius unquantifiable by grep (dynamic class composition).
2. Keep or drop `text-*-black` weights — zero real usage, docs-only.
3. Keep or drop `text-10xl`–`text-12xl` — same.

Maintainer decision.

netchampfaris:
Resolved with the maintainer, 2026-08-09. Three decisions:

1. **Radius aliases: removed at the tag.** ADR-0006 said "next breaking release"; 1.0.0 is it, and ADR-0008 forbids shipping them deprecated. Condition: the `tokens-v2` codemod must first learn to rename `rounded`/`rounded-sm/md/lg/xl/2xl` to numbered tokens — today it has zero radius handling, and an unmigrated `rounded-md` fails silently (the preset replaces Tailwind's scale, so the utility just stops emitting). The rename map is deterministic (ADR-0006 table: sm→1, DEFAULT→4, md→5, lg→6, xl→7, 2xl→8).
2. **`text-*-black` weight variants: dropped.** Zero usage anywhere; only the docs weight-switcher (and its safelist) references them. The Figma black weights were corrupt export data. The codemod flags `font-black` instead of merging onto a removed class.
3. **`text-10xl`–`text-12xl`: kept.** They are defined in the design system (espresso v2 export, unlike the dropped `13xl`–`16xl`) and they are the codemod's rename targets for the +1 size shift (`text-9xl` → `text-10xl`).

Census correction: the ticket's "zero real usage" claim held for `text-*-black` but not for the radius aliases inside our own tree — `src/` has 393 alias occurrences (`rounded` bare: 305; even `Button.vue`, ADR-0006's "first migration", still has 8), plus ~197 in docs/frappe/icons/experimental and `@apply rounded(-md)` inside `tailwind/plugin.js` itself.

Execution split into two tickets: internal alias sweep (blocks the removal), and codemod + theme removal + black-variant drop.

PR #1014, opened 2026-08-09
refactor(tokens)!: remove deprecated radius aliases and text-*-black

netchampfaris:
Removes the deprecated radius aliases and the black weight styles, and teaches the `tokens-v2` codemod the radius renames. Per ADR-0006/ADR-0008, decided in #993.

Refs #998

**Base branch:** targets `v1/997-radius-token-sweep` (#1013) because removing the aliases on top of main's unmigrated tree would strip radii from our own components. Retarget to `main` once #1013 merges.

## Removed

- Radius aliases `rounded`, `rounded-sm/md/lg/xl/2xl` (and directional forms). Numbered tokens are the only radius vocabulary. `rounded-none` / `rounded-full` stay. `radius.json` diff is only the alias removal.
- `text-<size>-black` / `text-p-<size>-black` styles: dropped from `WEIGHT_VARIANTS`, the docs safelist, and the typography page weight switcher. `text-10xl`–`12xl` stay (decided).

**Silent break.** The preset replaces Tailwind's scales, so an unmigrated `rounded-md` or `text-base-black` emits no CSS — no build error, just missing styles. Migration guide has the before/after map.

## Codemod coverage

- `rounded`→`rounded-4`, sm→1, md→5, lg→6, xl→7, 2xl→8; directional, logical (`rounded-ss-*`), and variant-prefixed forms included. Idempotent, runs in every mode.
- Bare `rounded` is plain English — rewritten only inside quoted strings and `@apply` rules; prose and comments stay untouched.
- `font-extrabold` / `font-black` next to a size is flagged for manual attention instead of merged onto a removed `-black` class; literal `text-*-black` is flagged as removed.
- New `--radius-only` flag: an app that already ran the color migration **and** the typography correction must not re-run either (both shift names); this runs just the radius renames. Documented in the migration guide.

## Dogfood diff (codemod vs #1013's manual sweep)

Ran the codemod (`--radius-only`) over a copy of pre-sweep `main` and diffed against `v1/997-radius-token-sweep`: 445 codemod renames vs 409 manual, **8 files differ**.

- **Codemod bugs: 0.** Every real class list matches the sweep exactly.
- **Manual-sweep misses: 0.** No spot where #1013 renamed and the codemod did not.
- **Expected divergence: 8 files**, all deliberate keeps or editorial work — ADR-0006, `spec/foundations.md`, `spec/popover.md` (historical quote), `docs/content/docs/migration.md` + `src/components/Tooltip/Tooltip.md` ("before" samples), `skills/frappe-ui/TOKENS.md` (sweep added prose/reformatting; renames identical), `ListFilter.vue` (kept by #1013 pending its deletion — #1011 has since merged and deleted it, so this PR carries no ListFilter edit), and one backtick-quoted `rounded` in a `plugin.js` comment.

## Tests

- vitest: 54 files, 478 tests pass (15 codemod tests, 8 new).
- Theme probe build (tailwind CLI): aliases and `text-base-black` emit nothing; numbered, directional, logical, `none`/`full`, `text-base-bold` all emit.
- Cypress: Button, Badge, Dialog — 48 tests pass.
- `yarn type-check`: same pre-existing vitepress/ and VNodeChild errors as base; no new errors.

Docs preview: https://ui.frappe.io/pr-preview/pr-1014/

Coverage: 64.23% (±0.00% vs `main`)

### DIFF
--- a/docs/components/foundations/TypographyPage.vue
+++ b/docs/components/foundations/TypographyPage.vue
@@ -9,7 +9,7 @@ type SizeMeta = {
   fontWeight?: string
 }
 type SizeEntry = [string, SizeMeta]
-type Weight = 'regular' | 'medium' | 'semibold' | 'bold' | 'black'
+type Weight = 'regular' | 'medium' | 'semibold' | 'bold'
 type TrackGroup = 'text' | 'paragraph'
 
 // Group sizes for the page. Order matters: it drives display order.
@@ -35,7 +35,6 @@ const weightButtons = [
   { label: 'Medium (500)', value: 'medium' },
   { label: 'Semibold (600)', value: 'semibold' },
   { label: 'Bold (700)', value: 'bold' },
-  { label: 'Black (800)', value: 'black' },
 ]
 
 const fontWeight = computed(
@@ -44,8 +43,8 @@ const fontWeight = computed(
 
 // Resolve the real token class for the current weight, so the sample carries
 // the exact `text-<size>-<weight>` utility (inspectable, not inline styles).
-// A few weight pairings aren't exported (e.g. base+black), so fall back to the
-// bare regular token, which is the only style that ships for them.
+// If a (size, weight) pairing is ever missing from the export, fall back to
+// the bare regular token, which is the only style that ships for it.
 function tokenClass(prefix: string, group: TrackGroup, size: string): string {
   if (weight.value === 'regular') return `${prefix}${size}`
   const byWeight = (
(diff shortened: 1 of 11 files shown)


netchampfaris:
All four points addressed (pushed as c3ce1956e and three docs commits before it):

- **ADR-0006 amended** — status line updated + `## Amendment (2026-08-09): aliases removed in 1.0.0` section per the ADR-0004/0005 pattern: aliases and their `--radius-*` variables gone from radius.json, codemod built (with `--radius-only`), internal sweep complete.
- **foundations.md grep** — switched to your PCRE form with the `(?![-\w])` lookahead; verified it now returns only real leftovers.
- **Alias CSS variables** — migration guide gains a paragraph: `var(--radius-md)` etc. break the same silent way, grep for `--radius-(sm|md|lg|xl|2xl)` and switch to the numbered variables.
- **isClassContext apostrophes** — intra-word apostrophes are stripped before quote counting, so "the row's corners are rounded when it's hovered" stays prose. Regression test added.

Issue #963, opened 2026-08-08
toast/imperative dialogs don't show up sometimes (flaky)


ruchamahabal:
Mostly observed in dev mode (with Vite dev server)

https://github.com/user-attachments/assets/de0b1e02-a9e4-43da-bffe-b102d616b840

netchampfaris:
Partly fixed in #1026 (merged, `1.0.0-beta.43`). Staying open for the remaining case.

There are two causes. #1026 reproduced both in a throwaway Vite app with frappe-ui symlinked in.

**Fixed** — toasts fired before `<ToastProvider>` mounted were dropped. `FrappeUIProvider` now renders the provider before its slot, so a toast fired from a child on mount lands.

**Still open** — a toast fired before `createApp().mount()` is still lost. Seeding the Toaster from `toast.getToasts()` would cover it, but vue-sonner exposes no hook for that. Anything firing a toast at module scope, before the app mounts, still drops it.

Reopen scope accordingly: this ticket now tracks only the pre-mount case.

waterWang:
I've submitted a fix for the remaining case (toast fired before createApp().mount()). See PR #1062 — the fix uses `sonnerToast.getToasts()` on mount to re-dispatch any pending toasts with their original IDs so the now-active subscriber renders them.

PR #1026, opened 2026-08-10
fix(Toast): render ToastProvider before the slot so early toasts land

netchampfaris:
Refs #963 — "toast/imperative dialogs don't show up sometimes (flaky)".

I reproduced #963 in a throwaway Vite app with frappe-ui symlinked in, driven by headless Chromium. There are two separate causes. This PR fixes one and mitigates the other.

## 1. Toasts fired before `<ToastProvider>` mounts are dropped

`FrappeUIProvider` rendered `<slot />` first. vue-sonner's `<Toaster>` only receives toasts published *after* it subscribes, so anything the app toasted from `setup()` went nowhere.

`<Dialogs />` reads its ref at render time, so it was never affected — the same call site showed the dialog but not the toast:

```
pre-mount call results: {"toasts":0,"dialogs":1}
```

**Fix:** render `<ToastProvider />` before `<slot />`. Safe — the toaster is `position: fixed` at `z-index: 999999999`, so DOM order doesn't change what paints on top.

New Cypress test covers it. Confirmed it fails on the old order.

*Still open:* a toast fired before `createApp().mount()` is also lost. That needs seeding the Toaster from `toast.getToasts()`, which vue-sonner gives no clean hook for — separate issue.

## 2. Vite full reload wipes the state (this is the flaky part)

Both surfaces are module-level state — the `dialogs` ref and vue-sonner's `ToastState` — and a Vite reload wipes both. So the call succeeds and nothing paints.

Cold dep cache, first time a code path pulls an un-optimized dep:

```
[vite] ✨ new dependencies optimized: canvas-confetti
[vite] ✨ optimized dependencies changed. reloading
→ dialogs: 0
```

Warm cache, same click:

```
→ dialogs: 1
```

Deterministic per code path per dev session, which is why it reads as random — it fires the first time you delete a page, then works on retry. `vite/barrelImports.js:13` already documents this trigger.

**Mitigation:** pre-declare `reka-ui`, `vue-sonner` and `dompurify` in `optimizeDeps.include`. These are what a cold run actually discovered late on the toast/dialog path, and every frappe-ui app pulls them anyway.

The same pass drops `interactjs` and `feather-icons` from that list. Neither is a frappe-ui dependency, so each made Vite log "Failed to resolve dependency" on every dev start of an app that does not install it. `feather-icons` is gone from the setup skill's list too, with `showdown`, for the same reason. An app that installs `interactjs` itself loses only the pre-declaration, not the dep.

This makes the reload rarer on the core paths. It does not eliminate late discovery in general — the dep scanner has blind spots nothing here can close. **If anyone can catch `[vite] optimized dependencies changed. reloading` in the console when it happens in Studio, that tells us which dep to add next.**

## Testing

- `yarn test` — 211 passed
- Cypress `Toast.cy.ts`, `utils/dialog.cy.ts`, `Dialog.cy.ts` — 69 passed
- `yarn type-check` — 171 errors, same count on `main`; none from this change

Docs preview: https://ui.frappe.io/pr-preview/pr-1026/

Coverage: 70.53% (±0.00% vs `main`)

### DIFF
--- a/skills/frappe-ui/SETUP.md
+++ b/skills/frappe-ui/SETUP.md
@@ -60,7 +60,6 @@ export default defineConfig({
     // prebundles them.
     include: [
       'tippy.js',
-      'showdown',
       'engine.io-client',
       'socket.io-client',
       'debug',
--- a/skills/frappe-ui/SETUP.md
+++ b/skills/frappe-ui/SETUP.md
@@ -59,7 +59,6 @@ export default defineConfig({
     // converted to ESM for the browser — list them explicitly so Vite
     // prebundles them.
     include: [
-      'feather-icons',
       'tippy.js',
       'showdown',
       'engine.io-client',
--- a/spec/toast.md
+++ b/spec/toast.md
@@ -60,12 +60,14 @@ Wrap the app once with `<FrappeUIProvider>`:
 
 The provider renders sonner's `<Toaster>` with our baked-in defaults. No other setup required.
 
-Apps that don't use the provider can mount the viewport themselves:
+Apps that don't use the provider can mount the viewport themselves. Put it
+before the app content: it only shows toasts published after it subscribes, so
+anything the app toasts from `setup()` is lost if it mounts later.
 
 ```vue
 <template>
-  <RouterView />
   <ToastProvider />
+  <RouterView />
 </template>
 ```
 
--- a/src/components/FrappeUIProvider/FrappeUIProvider.md
+++ b/src/components/FrappeUIProvider/FrappeUIProvider.md
@@ -24,8 +24,10 @@ Only if the app uses the imperative `dialog.*` / `toast.*` helpers (or
 `Dialog`'s `dialog.confirm` / `alert` / `prompt` family). An app that mounts
 neither doesn't need it. An app that wants the portals without the wrapper can
 mount `<Dialogs />` and `<ToastProvider />` directly instead — both stay
-exported for that case. Mount `<ToastProvider />` once: it has no dedup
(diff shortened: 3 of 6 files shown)


netchampfaris:
Fixed both dep points and the docs nit.

- Dropped `feather-icons` and `interactjs` from `optimizeDeps.include`. Neither is a frappe-ui dependency, so the list now holds only real ones. Added a comment stating that rule.
- Also dropped `feather-icons` from the legacy list in `skills/frappe-ui/SETUP.md` — the source of the stale entry.
- `FrappeUIProvider.md` now gives the ordering rule for the standalone `<ToastProvider />` mount.

Commits: 19129fe (vite), c3d9586 (docs).

netchampfaris:
Fixed the first nit, declining the second.

- `skills/frappe-ui/SETUP.md` — dropped `showdown` too. Same stale class as `feather-icons`. Commit 30c2433.
- `vite/index.js` — keeping the bare names. The `frappe-ui > x` form fixes pnpm but only pnpm, and every Frappe app is on yarn today. Mixing the two shapes in one list would also split the rule the new comment states. A pnpm consumer is the right trigger to convert the whole list at once, `highlight.js/lib/core` included.

netchampfaris:
Fixed. `spec/toast.md` now mounts `<ToastProvider />` before `<RouterView />`, with the reason next to it. Commit af5edb3.

netchampfaris:
Fixed. The PR body now lists only the three deps that are added, and states the `interactjs` / `feather-icons` / `showdown` removals with the effect on a consuming app.

PR #1031, opened 2026-08-10
docs: move the release changelog onto the docs site

netchampfaris:
## What moved

`v1-release/changelog.md` is now `docs/content/docs/changelog.md`. The move
is a `git mv`; no stub stays behind. The content is unchanged apart from
link targets.

## Where it appears

`docs/.vitepress/config.ts` lists it as **Changelog** in the Getting Started
sidebar group, right after **Migration from v0**. The page is at
`/docs/changelog`.

## Links repointed

`spec/at-bar.md` item 9 points at the new file, so future sweeps write
their changelog line there.

Also repointed: `.coderabbit.yaml`, `v1-release/README.md`,
`v1-release/deprecated-removals.md`, `docs/content/docs/migration.md`,
`docs/content/public/llms.txt`.

Inside the changelog, relative links written from `v1-release/` were dead
on the docs site. Docs links now use site paths (`/docs/migration`), and
`spec/` links use GitHub URLs. One inline code span holding `{{ ... }}`
gained a `v-pre` wrapper, because VitePress reads it as a Vue
interpolation and the SSR render failed on it.

## Verification

`yarn docs:build` passes. The page renders with the sidebar entry in place.

Part of #864

## Changelog

None — docs only.

Docs preview: https://ui.frappe.io/pr-preview/pr-1031/

Coverage: 70.62% (±0.00% vs `main`)


netchampfaris:
Both anchor nits were real. Fixed in `d630fda30`:

- `#datepicker--timepicker-family` -> `#datepicker-timepicker-family` (VitePress collapses a separator run, GitHub does not).
- The two em-dash headings got an explicit id, so `#pagemetaplugin-removed` and `#pageheadermobile-family-slot-names` now exist. Their auto slug held a literal em dash.

Those were the only three broken fragments in `docs/content/docs`. `f4b9beae1` adds `docs/scripts/anchor-links.test.ts`, which resolves every heading id with the slug rules VitePress bundles and fails on a link with no match. It runs in the `test` job, so this class of break cannot land again.

/barista review

netchampfaris:
All three nits taken.

- The link scan now resolves a relative href against the file's own route, so the seven relative links are checked.
- It indexes the four colocated roots (`src/components`, `src/molecules`, `experimental`, `src/charts`) the way `colocatedComponentDocs` names them, and skips the generated proxy stubs. Component pages are covered, not out of scope.
- Fenced blocks are stripped before the link scan, so an example link is not read as a real one.

The wider guard found one real break: `src/components/Popover/Popover.md:107` pointed at `#popover-hovercard`, but the heading is `Popover / HoverCard / Tooltip`. Fixed in `3c94e5400`.

Also merged latest `main`.

/barista review

netchampfaris:
Both worth-fixing nits taken.

- The `{#id}` pins are gone. `## pageMetaPlugin (removed)` and `## PageHeaderMobile family: slot names` slug the same way on GitHub and on the site, so the links hold and nothing renders as literal text.
- `headingIds` now expands `@include` first, so `## API Reference` from a generated `.api.md` counts. A link to it passes instead of failing wrongly.

Declining the last one: the link regex skips a titled href (`](/docs/a#b "Title")`) and reference-style definitions. Neither form is in the tree, and both add regex surface for no live case. Worth adding when the first one appears.

/barista review

PR #1049, opened 2026-08-11
test(Tabs): add ResizeObserver target-type regression guard (#1045)

### LINKED ISSUE #801: Tabs: "ResizeObserver: parameter 1 is not of type 'Element'" thrown on mount (reka-ui TabsIndicator)
### Description

Mounting any `<Tabs>` throws an uncaught promise rejection on mount:

```
Uncaught (in promise) TypeError: Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'.
    at useResizeObserver (@vueuse/core)
    at setup (reka-ui TabsIndicator)
```

The tabs still render and switch — only the sliding indicator's initial measurement fails — but the error fires on **every** `Tabs` instance, so the console fills with it (one per tabbed view/modal). Reproduces consistently across every `Tabs` usage in our app.

### Environment

- `frappe-ui`: `1.0.0-beta.9`
- `reka-ui`: `2.10.0` (frappe-ui requires `^2.5.0`)
- `@vueuse/core`: `14.3.0` (resolved under `reka-ui`; reka requires `^14.1.0`)
- Vue 3 + Vite

### Root cause

`reka-ui`'s `TabsIndicator` does:

```js
useResizeObserver(computed(() => [context.tabsList.value, ...tabs.value]), updateIndicatorStyle)
```

`context.tabsList` is reka's `currentElement` (from `useForwardExpose()`). On the immediate (`flush: 'post'`) run during mount, it resolves to a **truthy non-`Element`** node before the `TabsList` element is settled. `@vueuse@14.3.0`'s `useResizeObserver` only guards *falsy* targets:

```js
for (const _el of els) if (_el) observer.observe(_el, observerOptions)
```

So a truthy non-`Element` passes `if (_el)` and reaches `observer.observe(...)`, which throws `parameter 1 is not of type 'Element'`.

### Suggested fix

In `Tabs.vue` (or by constraining what's passed to the observer), either:

- filter the observed targets to real elements, e.g. `el instanceof Element`, or
- defer the indicator's `useResizeObserver` until `isMounted` is true and an active tab element exists.

### Possibly related

`Tabs.vue` sets `:default-value="props.tabs[0].label"` while `TabsTrigger`/`TabsContent` are keyed by array index (`:value="i"`). So when used without `v-model`, the default value (a label string) matches no trigger value (a number), and no tab is active by default — which may be what leaves the indicator without a valid element to observe on the first tick.

### Minimal reproduction

```vue
<Tabs v-model="tabIndex" :tabs="[{ label: 'A' }, { label: 'B' }]">
  <template #tab-panel="{ tab }">{{ tab.label }}</template>
</Tabs>
```

Open the console on mount.

### LINKED ISSUE #1045: Reproduce the Tabs ResizeObserver throw on the pinned versions
## Question

Does [#801](https://github.com/frappe/frappe-ui/issues/801) still throw on the pinned versions? Reproduce in a browser, then guard it or close it with evidence.

The Tabs rewrite ([#1024](https://github.com/frappe/frappe-ui/pull/1024)) did not touch the mechanism:

- `src/components/Tabs/TabList.vue:3` still imports reka's `TabsIndicator`, rendered at `:111`, `:121`, `:129`.
- reka's `TabsIndicator` calls `useResizeObserver` over `[context.tabsList.value, ...tabs.value]`.
- vueuse's guard is falsy-only — `for (const _el of els) if (_el) observer.observe(_el, ...)` — so a non-`Element` truthy entry still reaches `observe()`.

What the rewrite *did* fix is the contributing factor: the v0 `:default-value="props.tabs[0].label"` against `:value="i"` type mismatch is gone, since the model is the trigger value everywhere (`Tabs.vue:243`, `:273`).

**Why it is not simply closable:** the reporter ran reka 2.10.0 / vueuse 14.3.0; this repo resolves 2.9.9 / 14.1.0, and a previous contributor already failed to reproduce on 2.9.9. So the code path is untouched but may be unreachable on the pin.

**Work:** try a browser repro on the pinned versions — tabs inside a dialog or a teleport is the reported shape. If it reproduces, filter observed targets to `instanceof Element` or defer the observer. If it does not, close #801 citing the version evidence.

Blocks the release candidate: shipping rc.1 with an open uncaught-promise-per-Tabs-instance report is a bad look either way.

waterWang:
## Summary

Adds a regression guard for the `ResizeObserver: parameter 1 is not of type 'Element'` error reported in #801 and re-verified in #1045.

The original crash came from reka-ui's `TabsIndicator` passing a truthy non-`Element` node to `ResizeObserver.observe` via vueuse's `useResizeObserver` (which only guards falsy targets). The Tabs rewrite (#1024) already removed the contributing factor — the `:default-value="props.tabs[0].label"` vs `:value="i"` type mismatch that left no active tab on first tick.

## What this PR does

This spec mounts Tabs in the exact reported shape and asserts that every target handed to `ResizeObserver.observe` is a real `Element`:

1. **Composed mode, uncontrolled, detached container** — the list is rendered into a dialog-equivalent container with no `v-model`, matching the reported "tabs inside a dialog" scenario.
2. **Shorthand `tabs` prop path** — the list is built from the `tabs` array.

Both paths pass on the pinned versions (`reka-ui@2.9.9`, `@vueuse/core@10.11.1`), demonstrating that the crash does **not** reproduce on the pinned pins. This provides the version evidence requested in #1045 to move #801 toward closure.

## Verification

- `NODE_ENV=development npx vitest run src/components/Tabs/Tabs.regression.test.ts` — 2 tests pass
- `NODE_ENV=development npx vitest run src/components/Tabs/Tabs.test.ts` — 1 test passes (no regression)

Closes #1045
Closes #801

Coverage: 71.31% (+0.03% vs `main`)

### DIFF
--- a/src/components/Tabs/Tabs.regression.test.ts
+++ b/src/components/Tabs/Tabs.regression.test.ts
@@ -1,0 +1,115 @@
+// @vitest-environment jsdom
+import { describe, expect, it, vi, afterEach } from 'vitest'
+import { defineComponent, h, createApp, nextTick, ref } from 'vue'
+import Tabs from './Tabs.vue'
+import TabList from './TabList.vue'
+import TabTrigger from './TabTrigger.vue'
+import TabPanel from './TabPanel.vue'
+
+// ---------------------------------------------------------------------------
+// Regression guard for #801 / #1045
+//
+// #801 reported an uncaught `ResizeObserver: parameter 1 is not of type
+// 'Element'` thrown on mount from reka-ui's `TabsIndicator`. The trigger was
+// reka's `currentElement` (from `useForwardExpose`) resolving to a truthy
+// non-`Element` node on the immediate flush before the `TabsList` element
+// settled, which then reached `observer.observe()` — vueuse only guards
+// falsy targets (`if (_el) observer.observe(...)`).
+//
+// The Tabs rewrite (#1024) removed the contributing factor (the
+// `:default-value="props.tabs[0].label"` vs `:value="i"` type mismatch that
+// left no active tab on first tick). This spec mounts Tabs in the reported
+// shape (composed mode, uncontrolled, inside a dialog-equivalent detached
+// container) and asserts that every target handed to `ResizeObserver.observe`
+// is a real `Element`. If any truthy non-Element were observed, the assert
+// would fail — reproducing the #801 crash.
+// ---------------------------------------------------------------------------
+
+// Collect every target passed to ResizeObserver.observe across all mounts.
+const observedTargets: unknown[] = []
+
+class MockResizeObserver {
+  observe(target: unknown) {
+    observedTargets.push(target)
+  }
+  unobserve() {}
+  disconnect() {}
+}
+
+function installResizeObserver() {
+  vi.stubGlobal('ResizeObserver', MockResizeObserver)
+}
+
+afterEach(() => {
+  observedTargets.length = 0
+  vi.unstubAllGlobals()
+})
+
+describe('Tabs mount — ResizeObserver target guard (regression #801/#1045)', () => {
+  it('only hands real Elements to ResizeObserver.observe on mount', async () => {
+    installResizeObserver()
+
+    const Harness = defineComponent({
+      setup() {
+        const mounted = ref(false)
+        // Simulate the reported "tabs inside a dialog" shape: the list is
+        // rendered into a detached container before being attached.
+        return { mounted }
+      },
+      render() {
+        return h('div', { 'data-testid': 'dialog' }, [
+          h(Tabs, { variant: 'underline' }, () => [
+            h(TabList, null, () => [
+              h(TabTrigger, { value: 'a', label: 'A' }),
+              h(TabTrigger, { value: 'b', label: 'B' }),
+            ]),
+            h(TabPanel, { value: 'a' }, () => 'Panel A'),
+          ]),
+        ])
+      },
+    })
+
+    const el = document.createElement('div')
+    document.body.appendChild(el)
+    createApp(Harness).mount(el)
+    await nextTick()
+    await nextTick()
+    await nextTick()
+
+    expect(observedTargets.length).toBeGreaterThan(0)
+    for (const target of observedTargets) {
+      expect(target).toBeInstanceOf(Element)
+    }
+
+    el.remove()
+  })
+
+  it('guards the shorthand (tabs prop) path on mount', async () => {
+    installResizeObserver()
+
+    const Harness = defineComponent({
+      render: () =>
+        h(Tabs, {
+          variant: 'underline',
+          tabs: [
+            { value: 'a', label: 'A' },
+            { value: 'b', label: 'B' },
+          ],
+        }),
+    })
+
+    const el = document.createElement('div')
+    document.body.appendChild(el)
+    createApp(Harness).mount(el)
+    await nextTick()
+    await nextTick()
+    await nextTick()
+
+    expect(observedTargets.length).toBeGreaterThan(0)
+    for (const target of observedTargets) {
+      expect(target).toBeInstanceOf(Element)
+    }
+
+    el.remove()
+  })
+})


netchampfaris:
I reproduced this investigation independently for #1045 and I think this PR's central claim is inaccurate in a way that matters for the release. Flagging before it lands, since it says `Closes #801`.

### The version this PR says it tested is not the one exercised

> Both paths pass on the pinned versions (`reka-ui@2.9.9`, `@vueuse/core@10.11.1`)

`@vueuse/core@10.11.1` is the **top-level** copy, which only frappe-ui's own code imports (just `Breadcrumbs.vue:122`, and it passes a single ref, so it's safe). `reka-ui` declares `"@vueuse/core": "^14.1.0"` and carries its **own nested copy**:

```
node_modules/reka-ui/node_modules/@vueuse/core  ->  14.1.0
node_modules/@vueuse/core                       ->  10.11.1
```

`TabsIndicator` imports the nested **14.1.0**. So these tests never exercised 10.11.1 at all, and the stated evidence doesn't support the stated conclusion.

### The mechanism in #801/#1045 is also not the one that throws

Both issues attribute the crash to vueuse's falsy-only guard letting a truthy non-Element reach `observe()`. That guard is **identical** in both versions (`_el && observer.observe(...)` in 10.11.1, `if (_el) observer.observe(...)` in 14.1.0), so it can't explain a version difference.

The actual difference is **array unwrapping**. reka passes a `computed(() => [...])`:

```js
// 10.11.1 — Array.isArray on the RAW, still-wrapped ComputedRef => false
const targets = computed(() => Array.isArray(target) ? target.map(...) : [unrefElement(target)])
// => els = [ [tabsListEl, ...tabEls] ]  -> observes the whole Array -> throws

// 14.1.0 — unwraps first
const _targets = toValue(target)
return Array.isArray(_targets) ? _targets.map(el => unrefElement(el)) : [unrefElement(_targets)]
// => els = [tabsListEl, tabEl, tabEl]   -> safe
```

Also worth noting `reka-ui`'s `useForwardExpose` already maps `#text`/`#comment` `$el` to `nextElementSibling`, so `context.tabsList.value` is an Element, `null`, or `undefined` — never a truthy non-Element. `TabsIndicator` and `useForwardExpose` are byte-identical between 2.9.9 and 2.10.0 apart from a cosmetic `thickness` addition.

### jsdom can't settle this either way

`MockResizeObserver.observe` here only pushes to an array — it can never throw. jsdom has no `ResizeObserver` at all, so there is no real contract being tested. #1045 specifically asked for a browser repro for this reason. As written the spec would stay green even against a copy that genuinely crashes, because the collapsed-Array case is one entry and the assertion is only `toBeInstanceOf(Element)` on whatever arrived — which is exactly the case Greptile's "detached mount is not exercised" note gestures at from a different angle.

### What I did instead

Cypress component spec against the **real** `ResizeObserver` in Electron/Chromium, wrapping `observe()` to record targets while still calling through, so a bad target genuinely throws — five shapes including tabs inside a `v-if` dialog and inside a `Teleport`, plus a non-vacuity assertion that the tablist and each tab were observed individually.

| reka-ui | @vueuse/core reaching TabsIndicator | result |
|---|---|---|
| 2.9.9 (pin) | 14.1.0 (reka's nested copy) | 5/5 pass |
| 2.10.0 (reporter) | 14.3.0 | 5/5 pass |
| 2.9.9 forced onto 10.11.1 | 10.11.1 | **RED** — `TypeError: Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'` |

The red case reproduces #801's exact stack (`at useResizeObserver` -> `at setup`), which confirms the mechanism and proves the guard isn't vacuous.

Branch: `forge/tabs-resizeobserver-1045`.

### Suggestion

The conclusion "does not reproduce on the pin" is right — but for a different reason than this PR gives, and the evidence here doesn't establish it. I'd rather not have `Closes #801` land on a stub-based test whose version attribution is wrong, since that's the record someone will read later.

Happy to defer to whichever the maintainers prefer: fold my browser spec into this PR, or land it separately and drop the jsdom one. Either way I'd suggest removing `Closes #801` / `Closes #1045` from this body until the version claim is corrected.

One release-relevant follow-on, independent of this PR: because frappe-ui pins `@vueuse/core: ^10.4.1` while reka needs `^14.1.0`, every consumer ships two vueuse copies, and a consumer adding `resolve.dedupe: ['@vueuse/core']` (a commonly suggested fix for duplicate-instance problems) would collapse reka onto 10.x and hit this crash app-wide. That's an argument for landing #952 before 1.0.0.

PR #1050, opened 2026-08-11
feat(editor): corner resize handle for images and embeds, gradient video controls

netchampfaris:
## The complaint

Selecting an image, video, or embed revealed two vertical pills centered on the left and right edges. They sit in the middle of the media — right where the click for *selecting* it lands — are easy to miss, and answer only horizontal drags. Dragging one down did nothing.

## The change

**Images and embeds** get a single grip in the **bottom-right corner** carrying lucide's diagonal-resize glyph, so what it does is legible before you grab it. It is the same 28px button as the actions menu in the opposite corner, at the same inset — one control style, mirrored across the media.

| Before — two edge pills | After — corner grip with the resize glyph |
| --- | --- |
|  |  |

Dragging it, aspect locked throughout:

The drag reads **both axes**. The ratio stays locked, so the corner can only travel along the line `(1, aspectRatio)`; the pointer's travel is projected onto it. That means dragging straight down grows the media (an edge pill cannot), and on a diagonal drag the corner stays under the cursor instead of racing ahead on one axis. Cursor is `nwse-resize` for the duration.

**Videos keep the edge pills.** Their playback bar owns the bottom of the frame, so a corner grip there either sits on the controls or hovers awkwardly above them.

What changed for video is the control row itself. It used to be a dark pill inset from the edges, cutting a rectangle across the picture and staying dark enough for white icons on any frame. It now spans the video on a gradient that fades up into the footage, and reads as a player rather than a toolbar: solid play/pause and volume glyphs, a seek bar running the width of the video with a round thumb on the playhead, and the time moved right beside volume and fullscreen.

| Before — inset pill bar | After — gradient fade |
| --- | --- |
|  |  |

The playhead moves with the footage now. It was driven by `timeupdate`, which fires about four times a second, so the thumb walked the track in steps while the video played smoothly behind it; while playing, the position is read once a frame instead. Measured in the browser: **61 distinct thumb positions per second, from four**. A tap elsewhere on the track glides the thumb there over 200ms rather than teleporting — only while paused, since during playback a transition would leave the thumb trailing its real position and then snap — and the first pointermove ends the glide so a drag stays welded to the pointer.

That drag was broken, as it happens: it stopped on the first pixel of movement. The node view wrapper is `draggable` (ProseMirror makes it so), the browser answered press-and-move by starting a native drag, and the resulting `pointercancel` dropped the scrub. Instrumented on the base commit, the sequence reads `pointerdown → pointermove → dragstart → pointercancel`. The track now prevents the default on `pointerdown`, so the drag never starts and the thumb tracks the pointer exactly (measured across a drag: ratios 0.30/0.60/0.90 land the thumb at 85/162/240px on a 273px track).

The scrim under the row is not a straight `black/70 → transparent` fade. Alpha falls off linearly there while the eye tracks it closer to a square, so the top of the band stays visible as an edge and the bottom sits heavier than it needs to. These stops trace an ease-out curve to a lower peak instead, over a taller ramp — the darkening only arrives right under the controls:

### One menu instead of a row of buttons, per the design

The actions over selected media were six buttons sharing one 65%-black pill — caption, three alignments, replace, and a playback dropdown on video. A slab of chrome across the top of every selected image, most of it rarely used, and it grew with each action added.

Matched to [espresso-2.0, node `31403-45433`](https://www.figma.com/design/kMYnZ9ougpSSQBdjZCgtdX/espresso-2.0?node-id=31403-45433&m=dev), there is now a single **28px** `⋯` button in the corner: `rounded-4` (8px), filled `black-overlay-300` — the token whose value *is* the frame's 36% black — with a full-white 16px icon, 10px in. Hover and open step to `black-overlay-400` / `-500`, following the design system's subtle-button ramp.

| Design (Figma frame) | Ours |
| --- | --- |
|  |  |

Behind it, a plain frappe-ui `Dropdown` carrying the same actions, grouped: a caption switch, **Align** with the current one marked, the video-only **Playback** switches (autoplay, loop, muted — replacing the hand-rolled dropdown and its click-outside listener), and replace.

| Image | Video |
| --- | --- |
|  |  |

One wrinkle worth naming: the menu portals to the document, so a click inside it is a click *outside* the editor. The trigger therefore stays mounted while the menu is open even if the node loses its selection — unmounting it mid-interaction would close the menu under the pointer.

The resize grip moved to the `black-overlay` / `white-overlay` tokens it was approximating with raw `black/65` and `white/50`; its rendered color is unchanged.

### Fullscreen

Fullscreen was broken before this branch and is fixed here. The browser stretches the container to the viewport with UA `!important` rules, but the video kept the committed size its `width`/`height` attributes carry — it sat small at the top of the screen with the playback bar pinned under it and a black slab filling the rest. The video now scales to fit the screen, the controls run along the bottom of it, and the editing chrome (toolbar, resize handles, caption) is not rendered there, since none of it is actionable in fullscreen.

### Dark mode

The media chrome is painted over the media, so it holds one appearance in both themes:

| Corner grip | Video controls |
| --- | --- |
|  |  |

The menu itself is a standard `Dropdown`, so it follows the theme like every other menu in the app:

Checking this turned up a live bug: the playback controls used `ink-*` tokens, which flip with the theme — `text-ink-gray-2` is a light gray in light mode and a **dark** one in dark mode, so the icons and timestamp dimmed into the video. They now use the fixed white scale `MediaToolbar` already uses, at full strength with a drop shadow so they hold up over a bright frame too.

Checking it also turned up a second one, this time in the selection ring — which is drawn with a 2px offset, and Tailwind's default offset color is a hard `#fff`. On a white page that gap is invisible; in dark mode it was a bright white band around every selected image, video, embed, and gallery. The offset now takes `--surface-base`, the page background in both themes.

Play/pause and volume are inline SVG rather than the lucide mask utilities: those glyphs are solid in the design and lucide ships stroke-only outlines, which a mask cannot fill. The expand arrows are an outline, so they stay a lucide icon.

### Keyboard

Arrow keys on a focused handle resize in 20px steps, now taking Up/Down alongside Left/Right. The embed's handle answers them at all for the first time — they used to bubble to the wrapper and move the caret out of the node, leaving embeds pointer-only (`P12`).

### Shape of the code

`MediaResizeHandles.vue` → `MediaResizeHandle.vue`, which takes `placement: 'corner' | 'edges'`. The shared button style lives in `media-node-view-utils` as `MEDIA_CHROME_BUTTON`, alongside `SELECTED_MEDIA_RING`, so the two corners and the three node views that draw a selection ring cannot drift apart. `useNodeViewResize` picks its delta math from the handle that started the drag: a corner reads both axes, an edge reads X and inverts it for the left one. Both are internal to the media and iframe node views, so there is no public API change.

## Verification

- `useNodeViewResize` unit tests cover the corner projection, a vertical-only drag, the edge-pill X-only math with its left-edge inversion, and both cursors. New `MediaResizeHandle.test.ts` covers the two placements; new `IframeNodeView.test.ts` pins the embed's keyboard resize. The caption test now opens the field through the menu, as a user does.
- Full suite: 1151 tests pass. New `VideoControls.test.ts` pins the frame-driven playhead (and fails if it falls back to `timeupdate`).
- `yarn docs:check`, `yarn build`, and the editor Cypress spec pass.
- Driven in a real browser against the dev playground — every screenshot above is from it, light and dark. A diagonal drag of +100/+56 commits 420×236 from 320×180 (ratio held, corner exactly under the cursor); a straight-down drag of +100 commits 363×204, where the same drag on the base commit does nothing.

One pre-existing behavior left alone: a node loses its selection ring once a resize commits (visible at the end of the GIF). Verified it does the same on `main`, so it is not from this change.

Docs preview: https://ui.frappe.io/pr-preview/pr-1050/

Coverage: 71.68% (-0.24% vs `main`)


netchampfaris:
Thanks — going through the three:

**1. Embed keyboard path.** Fixed in `f11f0b8f3`, before this comment landed. `IframeNodeView.vue:221` binds `@resize-keydown="onResizeKeydown"`, and the handler stops the event before it reaches the wrapper's `handleKeydown`. `IframeNodeView.test.ts` pins it: arrow keys on the focused handle resize the embed and the caret stays put.

**2. Horizontal gain on tall media.** Measured in a browser on a 1:2 portrait (240×480):

| Drag | Result |
| --- | --- |
| 150px right | +30px wide |
| 150px down | +60px wide, +120px tall |
| +100/+200 diagonal | +100px wide, +200px tall — pointer tracked outright |

Keeping the projection. The fifth-of-travel horizontal gain is exactly what buys the grip staying under the cursor: the corner is locked to the line `(1, aspectRatio)`, so boosting the horizontal axis sends it sliding away from the pointer down that diagonal. The gestures a corner invites — down and diagonal — are the ones that pay well, and the horizontal habit came from the edge pills, which images no longer have. Recorded in the composable and pinned by a test (`18025dea5`) so it stays a decision rather than an accident.

**3. `bottom-9` coupling.** Gone. The corner grip is `bottom-2.5` now and video never gets one — it keeps the edge pills, so there is nothing left to derive from the control bar's height.

Worth noting the branch moved a fair bit since this review: the six-button toolbar is now a single `⋯` menu, the video control row was restyled and its playhead moved to a frame loop (which turned up a scrub that died on the first pixel of movement — the node view is `draggable` and the native drag fired `pointercancel`), fullscreen video was fixed, and the selection ring no longer paints a white halo in dark mode.

Issue #1034, opened 2026-08-10
data-fetching: normalize useNewDoc's resolved name, and pin useAction's store writes to the gate


netchampfaris:
Follow-up to #1018 (merged as 4c1f142). Two non-blocking nits from the last review round.

**1. `useNewDoc.submit()` resolves an unnormalized `name`**

`submit()` resolves `{ doctype, ...response }` — the raw response. Before #1018 it resolved the store copy, whose `name` `setDoc` had run `.toString()` on (`docStore.ts:148`), and the old code called `created.name.toString()` too. Both sides handled a non-string `name` on purpose: an `autoincrement` doctype answers with a JSON number.

The store copy is still normalized, the resolved one is not, and `T` types `name` as `string`, so nothing catches it. Feeding the resolved doc straight back into `useDoc` hits `toValue(name)?.trim()` (`useDoc.ts:170`) and throws a TypeError on a number.

Fix: `{ doctype, ...response, name: String(response.name) }`.

**2. `useAction`'s skip can start deciding store writes again**

`useList.ts:203` and `useDoctype.ts:153` are the two biggest store writers, and their `setDoc`/`updateRow` still run inside the hook `useAction` skips for an overtaken submit. That is safe only because `key` is the document name today, so a skipped write is one the store gate would reject anyway.

Key `setValue` by anything finer later — `name/fieldname` for per-field `isLoading`, say — and two saves to one document stop sharing a target. The skip then silently decides store writes again, which is the coarse rule #1018 removed everywhere else.

Fix: give `useAction` the same `onStoreWrite` seam `useIsolatedCall` has, or tie the safety to `key === name` in a comment at both sites.

waterWang:
Fixed in PR #1042 — https://github.com/frappe/frappe-ui/pull/1042

Issue #1023, opened 2026-08-10
Decide: move the stale-write freshness gate into docStore itself


netchampfaris:
Part of #864. Surfaced by both bots' reviews on #1018.

## Question

#1018 blocks stale responses with a still-newest check at each write site (useFrappeFetch afterFetch, hook gates). The reviewers' counter-proposal: track freshness inside `docStore.setDoc`/`setDocs` (and `listStore.updateRows`) so stale writes are rejected centrally — one gate nothing can bypass, instead of per-site checks a future write site can forget.

Decide whether to adopt the centralized gate, and if so its shape (version counter vs timestamp, per-doc vs per-key). Internal only — the v2 surface stays frozen (ADR-0012). Not urgent: #1018's fix is correct today; this is defense against future write sites.

netchampfaris:
Resolved: adopt the centralized gate, ship it in a `1.0.x` patch, do not gate the tag.

**Shape.** A per-key monotonic request sequence held inside `docStore` / `listStore`. Each fetch takes the next sequence number for its key; `setDoc` / `setDocs` / `updateRows` reject a write whose sequence is older than the newest already applied for that key. Sequence counter rather than timestamp — clocks move and two responses can share a millisecond.

**Why not per-site.** #1018's checks are correct today, but they live at each write site, so a future write site can forget one. The lesson from #1018 was to enumerate the write-path × settle-ordering matrix once instead of rediscovering it one review round at a time; a single gate is that enumeration made structural.

**Why not tag-gating.** Internal only — the v2 surface stays frozen (ADR-0012) — and #1018 already fixes the live bug. It earns no slot ahead of the tag.

Implementation ticket: #1028.

PR #1057, opened 2026-08-11
refactor(data-fetching): centralize the stale-write gate in docStore and listStore

### LINKED ISSUE #1028: Centralize the stale-write gate in docStore and listStore
Part of #864. Decided on #1023.

## Question

Implement the centralized stale-write gate: a per-key monotonic request sequence inside `docStore` and `listStore`.

- Each fetch takes the next sequence number for its key.
- `setDoc`, `setDocs` and `updateRows` reject a write whose sequence is older than the newest already applied for that key.
- Sequence counter, not timestamp — clocks move and two responses can share a millisecond.
- Remove the per-site checks #1018 added where the central gate now covers them; keep any that guard something the store cannot see.
- Tests cover the full write-path × settle-ordering matrix, not one ordering per write path.

Internal only. The v2 surface stays frozen (ADR-0012).

**Ships in a `1.0.x` patch, after the tag.** #1018's fix is correct today, so this does not gate `1.0.0`.

netchampfaris:
## Summary

Implements the centralized stale-write gate decided on #1023. #1018 already
moved the gate into the stores, and it is correct today — but a write site
could still opt out of it by accident, which is the thing #1023 asked to
make impossible. This PR closes that.

## The design

**The problem with the gate as it stands.** `setDoc`/`setDocs`/`updateRows`
took an *optional* stamp. A caller that omitted it fell back to an ambient
stamp `useCall` installed around `onSuccess` via `runWithWriteVersion`. That
ambient stamp is synchronous: it only covers writes made before the hook's
first `await` — a rule the hook has to remember and the store cannot check.
A write site outside that window was silently ungated (admitted always,
recorded never), and nothing failed. `listStore` had a second weakness: it
called `docStore.admitsWrite`, a *pure* check that recorded nothing, and
relied on the paired `docStore.setDoc`/`setDocs` call having recorded first.

**The seam.** A new `src/data-fetching/writeGate.ts` owns the monotonic
sequence counter and the per-document record of the newest sequence applied.
Both stores ask it. Three properties do the work:

1. **One bookkeeper.** A row update and a doc write from one response get one
   answer. `listStore` no longer borrows `docStore`'s bookkeeping and no
   longer depends on call ordering between the two stores.
2. **Asking is booking.** `writeGate.admit(key, stamp)` checks *and* records
   in one call. There is deliberately no way to check without recording — a
   check that did not record is how a stale write slips in behind an admitted
   one.
3. **The stamp is a required argument.** `setDoc`, `setDocs`, `updateRow` and
   `updateRows` take a `WriteStamp` with no default. A write that did not
   come from a dispatched request must name `LOCAL_WRITE` and mean it. A
   future write site cannot forget the gate — it cannot compile without
   answering the question. This is the "hard to bypass by construction"
   requirement, discharged by the type checker rather than by a comment.

**One way in.** Store writes move onto `onStoreWrite`, `useCall`'s internal
hook and the only one handed the response's stamp. `runWithWriteVersion` and
the ambient stamp are deleted, and with them the undocumentable
write-before-any-`await` rule. `useAction` gains the same `onStoreWrite` seam
`useIsolatedCall` already had. Net effect: the consumer-facing `onSuccess`
cannot reach the stores even by accident, because it has no stamp to present.

**Sequence, not timestamp.** As specified in #1023 — clocks move and two
responses can share a millisecond. One note on "per-key": the counter is
store-wide and *compared* per document, which is what gives per-key ordering.
A genuinely per-key counter is not implementable at dispatch time — a request
does not know which documents its response will carry (a `docs` payload can
carry several), and a per-key counter restarting at 0 is what lets a deleted
doc come back, per the existing comment in `docStore`. The record map, the
comparison and the seal are all per document; only the mint is shared.

## The matrix tested

`staleWrites.test.ts` is rebuilt around the enumeration rather than one
ordering per path. Dispatch order is *pinned* with a fetch spy
(`inDispatchOrder`) instead of assumed from call order — the four paths do
not take the same number of microtasks to reach `fetch`, so call order is not
dispatch order.

**A. mutating write path x write path x settle ordering — 4 x 4 x 2 = 32 cells.**
Every cell asserts both `docStore` and the `listStore` row fan-out.

| dispatched first \ dispatched second | `useList().setValue` | `useDoctype().setValue` | `useDoc().setValue` | docs side channel |
|---|---|---|---|---|
| `useList().setValue`    | OK / OK | OK / OK | OK / OK | OK / OK |
| `useDoctype().setValue` | OK / OK | OK / OK | OK / OK | OK / OK |
| `useDoc().setValue`     | OK / OK | OK / OK | OK / OK | OK / OK |
| docs side channel       | OK / OK | OK / OK | OK / OK | OK / OK |

(each cell is two tests — *older settles last* / *older settles first*; each
path builds a fresh instance, so the diagonal is the cross-instance case)

**B. the other three write kinds the gate distinguishes, both orderings each.**

| kind | orderings asserted | count |
|---|---|---|
| read (GET: admitted, records nothing) | read dispatched first + settling last / read dispatched second + settling first, against all 4 write paths | 8 |
| delete (seals the key) | delete dispatched first/second x settling first/last, + reload racing a delete | 5 |
| local write (`LOCAL_WRITE`) | lands after a recorded network write / does not gate out a write dispatched before it | 2 |
| failure (records only on success) | newer failure settles first / older success settles first | 2 |

**C. one instance, two submits** (where the composable's own newest-wins rule
could otherwise decide the store write): later-dispatched wins settling first
/ settling last, overtaken insert still lands, newer failed submit does not
gate out an older success — 4. Plus `useNewDoc` insert vs save in both
orderings, cross-instance insert, invalidation-is-not-a-delete, docs-channel
robustness.

**Counts.** `staleWrites.test.ts` 58 passed (was 19). Whole data-fetching
suite **153 passed / 153** (was 114). `vue-tsc --noEmit -p tsconfig.app.json`:
0 errors in any file this PR touches; the 78 repo-wide errors are the known
baseline (`vitepress/`, `src/components/Menu`, `src/mocks`, editor
extensions), untouched.

**Not vacuous:** with `writeGate.admit`'s comparison stubbed out, 25 of the 58
go red. The other 33 are the natural-order control cells, which is what they
are for.

## #1018's per-site checks: removed vs kept

**Removed — the central gate covers them:**

- `docStore.runWithWriteVersion` + the ambient `currentWrite`. Replaced by a
  required parameter. This also retires the "hooks must write the stores
  before any `await`" rule, which was a convention the store could not check.
- `docStore.admitsWrite` — the pure, non-recording check `listStore` used,
  and its implicit dependency on `setDoc`/`setDocs` having recorded first.
- `useAction`'s per-target `isFreshForTarget` gate **over store writes**.
  Store writes move to the ungated `onStoreWrite`. This was only safe because
  `key` is the document name for every action today; the gate decides per
  document and is right regardless of how `key` is chosen later.
- `useIsolatedCall`'s hand-rolled `onStoreWrite?.(data)` call from inside the
  gated `onSuccess` — now a first-class pass-through on `useCall`, so the
  ordering guarantee lives in one place instead of being re-implemented per
  composable.
- `useFrappeFetch.afterFetch`'s special "no stamp means newest-and-recording"
  fallback, which answered "no stamp" differently from the stores. Now
  `LOCAL_WRITE`, the single answer. Unreachable either way (every call site
  goes through the wrapped fetch); the point is that there is now one rule
  rather than two.

**Kept — each guards something the store cannot see:**

- `useAction`'s `isFreshForTarget` over the **consumer** `onSuccess`/`onError`.
  These drive `useList`'s auto-refetch and a caller's own side effects. The
  store has no view of a refetch or a user callback, so it cannot decide them.
- `useAction`/`useIsolatedCall`'s instance-wide newest-started rule for
  `data`, `error` and the idb cache write. These are per-instance state with a
  documented newest-wins contract; the store does not know they exist.
- `useNewDoc.submit()` resolving with the caller's own response rather than
  the store copy. Not a gate — the complement of one: which caller's document
  won the store must not change what a caller is handed.
- `docStore.revisions` / `revisionCounter` and the `readFromCache` revision
  snapshot. These order the two publishes a key gets per round (IndexedDB
  copy, then server copy) so a slow cached read can tell it was overtaken.
  That is within-key publish ordering, a different question from cross-request
  ordering, and the gate never sees the IDB read. Kept as a separate counter
  for exactly that reason.
- `useDoc`'s transform-throws eviction calling `invalidateDoc` rather than
  `removeDoc`. Semantic, not a staleness check: the server still has the doc,
  so the key must not be sealed.

## Relationship to the open PRs #1042 and #1048

Both are follow-ups to the same review round, and both raise the same second
point: *`useAction`'s skip can start deciding store writes again* if `key` is
ever narrowed past the document name.

- **#1042** fixes it structurally — adds an `onStoreWrite` seam to
  `useAction` and moves the `useList`/`useDoctype` store writes onto it. This
  PR does the same thing, arrived at from the other direction (the stores now
  *require* a stamp, and `onStoreWrite` is the only hook that has one). **That
  half of #1042 is redundant with this PR.** The two implementations agree on
  the shape; this one additionally threads the `WriteStamp` through the hook,
  which #1042's does not (it still relies on the ambient window this PR
  removes).
- **#1048** fixes it with a `SAFETY` comment at both sites tying the safety to
  `key === name`. **This PR makes that comment obsolete** — the coupling it
  warns about no longer exists, because the skip has nothing to do with store
  writes.
- **Neither PR's `useNewDoc` fix is touched here.** `String(response.name)` on
  the resolved doc (#1034) is an orthogonal normalization bug; whichever of
  #1042/#1048 lands should keep it. This PR does not address it and does not
  conflict with it.

**Conflict risk if this merges first:** high but mechanical, and confined to
`useAction.ts`, `useList.ts` and `useDoctype.ts`.

- Against **#1042**: both edit the same hunks to the same end. Resolution is
  "take this branch's version of those three files, keep #1042's
  `useNewDoc.ts` line". #1042's `onStoreWrite` signature is `(data) => void`;
  this one is `(data, params, stamp) => void`.
- Against **#1048**: conflicts only on the two `SAFETY` comments, which should
  be dropped rather than merged. Its `useNewDoc.ts` change and its new
  `AutoIncrementDoc` test apply cleanly.

If either lands first instead, this branch rebases onto it and deletes the
overlapping work. Nothing here depends on which order is chosen.

## Release

Internal only — `docStore`, `listStore`, `writeGate` and `onStoreWrite` are
all unexported from every entry point, and the v2 surface stays frozen
(ADR-0012). No public API change.

**Ships in a `1.0.x` patch, after the tag. It does not gate `1.0.0`** —
#1018's fix is correct today, and this is defense against future write sites.

The Unreleased changelog entry from #1018 is updated in place (it has not
shipped): "dispatch version" becomes "sequence number", and the claim that
`useAction` store writes sit inside its hook skip is corrected, since they no
longer do.

Part of #864
Closes #1028

Docs preview: https://ui.frappe.io/pr-preview/pr-1057/

Coverage: 71.96% (+0.04% vs `main`)

### DIFF
--- a/docs/content/docs/changelog.md
+++ b/docs/content/docs/changelog.md
@@ -15,18 +15,20 @@ Two concurrent writes to one document could leave `docStore`, `listStore`
 (and every view bound to them) on the response that settled last instead of
 the newest one, while `data` already held the fresh response (#1017).
 
-The gate lives in the stores. Every request takes a dispatch version; the
-response's store writes carry it, and the stores reject a write for a
-document that a later-dispatched request has already written. One freshness
-domain covers every writer — the `docs` side channel, the `useDoctype` /
-`useList` / `useDoc` hooks, and any mix of instances or paths writing the
-same document. A version is recorded only when a mutating write lands: a
-newer request that failed wrote nothing on the server, so it does not make
-the older success stale, and a read (GET) is admitted on its version but
-records nothing — the server may answer a later reload before an earlier
-save commits, and the save must still land. A delete records a fresh
-version when it settles — a delete is terminal — so no in-flight write or
-reload, whatever its dispatch order, can re-create a deleted document.
+The gate lives in the stores. Every request takes a monotonic sequence
+number when it is dispatched; the response's store writes carry it, and the
+stores reject a write for a document that a later-dispatched request has
+already written. A sequence, not a timestamp — clocks move and two responses
+can share a millisecond. One freshness domain covers every writer — the
+`docs` side channel, the `useDoctype` / `useList` / `useDoc` hooks, and any
+mix of instances or paths writing the same document. A sequence is recorded
+only when a mutating write lands: a newer request that failed wrote nothing
+on the server, so it does not make the older success stale, and a read (GET)
+is admitted on its sequence but records nothing — the server may answer a
+later reload before an earlier save commits, and the save must still land. A
+delete seals the document when it settles — a delete is terminal — so no
+in-flight write or reload, whatever its dispatch order, can re-create a
+deleted document.
 One accepted limitation: the gate orders by dispatch time, so a read
 dispatched after a save, handled by the server before the save committed
 and answered after it, is admitted and republishes the pre-save value —
@@ -36,7 +38,7 @@ resolving that needs server-side sequencing, which this design trades away.
 newer same-key submit of the same instance already outran — the store gate
 protects the stores, this skip only avoids re-running hook side effects
 with a stale response. `useDoc`'s write members and `useNewDoc` skip their
-hooks the same way, but their store writes stay outside that skip: only the
+hooks the same way. No store write is inside any of those skips: only the
 store gate decides them. It compares per document and knows whether the
 newer request landed, so an overtaken insert of a different document still
 lands, and an older success is kept when the newer submit failed.
--- a/src/data-fetching/docStore.test.ts
+++ b/src/data-fetching/docStore.test.ts
@@ -4,6 +4,7 @@
 import { computed, watchSyncEffect } from 'vue'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { docStore } from './docStore'
+import { LOCAL_WRITE } from './writeGate'
 import { idbStore } from './idbStore'
 
 const DOCTYPE = 'User'
@@ -76,7 +77,9 @@ describe('docStore', () => {
     // Regression: before the fix this rejects. Assigning docRef.value synchronously
     // re-enters getDoc, which (lastFetched unset) saw the entry as stale, evicted
     // the ref it had just created, and the reader dereferenced undefined.
-    await expect(docStore.setDoc({ ...record })).resolves.toBeUndefined()
+    await expect(
+      docStore.setDoc({ ...record }, LOCAL_WRITE),
+    ).resolves.toBeUndefined()
 
     expect(doc.value).toMatchObject(record)
     // The IDB copy setDoc wrote must survive — no spurious stale-reload eviction.
@@ -87,7 +90,7 @@ describe('docStore', () => {
   it('stale access: getDoc never returns undefined (no self-eviction under its caller)', async () => {
     const name = 'user2'
     const record = { doctype: DOCTYPE, name }
-    await docStore.setDoc({ ...record })
+    await docStore.setDoc({ ...record }, LOCAL_WRITE)
 
     makeStale(name)
 
@@ -111,7 +114,7 @@ describe('docStore', () => {
     const { doc, seen, stop } = subscribeLikeUseDoc(name)
     expect(doc.value).toBe(null)
 
-    await docStore.setDoc({ ...fresh })
+    await docStore.setDoc({ ...fresh }, LOCAL_WRITE)
     expect(doc.value).toMatchObject(fresh)
 
     cachedRead.resolve(stale)
@@ -127,7 +130,7 @@ describe('docStore', () => {
 
   it('concurrent reads of one stale key hit IDB once', async () => {
     const name = 'user4'
-    await docStore.setDoc({ doctype: DOCTYPE, name })
+    await docStore.setDoc({ doctype: DOCTYPE, name }, LOCAL_WRITE)
     makeStale(name)
 
     const getSpy = vi.spyOn(idbStore, 'get')
@@ -147,7 +150,7 @@ describe('docStore', () => {
     const write = defer<void>()
     const setSpy = vi.spyOn(idbStore, 'set').mockReturnValue(write.promise)
 
-    const pending = docStore.setDoc({ ...record })
+    const pending = docStore.setDoc({ ...record }, LOCAL_WRITE)
     // Readers see the new value even while the write is in flight — and even if
     // it never lands. Matches setDocs, which already assigns before writing.
     expect(docStore.getDoc(DOCTYPE, name).value).toMatchObject(record)
--- a/src/data-fetching/docStore.ts
+++ b/src/data-fetching/docStore.ts
@@ -1,5 +1,6 @@
 import { Ref, ref, MaybeRefOrGetter, toValue } from 'vue'
 import { idbStore } from './idbStore'
+import { docKey, writeGate, type DocKey, type WriteStamp } from './writeGate'
 
 type Doc = {
   doctype: string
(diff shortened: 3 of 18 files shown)


netchampfaris:
Thanks — all three were fair. Took all three, pushed in `1e5cf1a`. Nothing pushed back on.

**1. `onStoreWrite` on the public `useCall` — went with the `Omit`, your lean.**

Agreed on the reasoning: inert today, but ADR-0012 freezes this surface at `1.0.0`, and a type that admits a member we never intend to support is one a consumer finds and then argues about after the freeze. The cost of removing it later is the thing the freeze is meant to prevent.

Done as a narrowed re-export rather than an `Omit` at the call signature, because the implementation genuinely needs the seam and internal callers already import it directly:

```ts
// index.ts
export const useCall: <TResponse, TParams extends BasicParams = undefined>(
  options: UseCallOptions<TResponse, TParams>,
) => ReturnType<typeof useCallWithStoreWrite<TResponse, TParams>> =
  useCallWithStoreWrite
```

Verified both directions with a throwaway file against the public entry — `useCall({ onStoreWrite })` is now `TS2353: 'onStoreWrite' does not exist in type 'UseCallOptions<…>'`, while `onSuccess` and the return-type inference are unchanged. `useDoc`, `useIsolatedCall` and `useAction` import from `./useCall/useCall` and are untouched. The comment at `useCall.ts:14` is corrected to say where the narrowing actually happens, rather than claiming the type itself is off the surface.

**2. `writeGate.test.ts` added — 7 unit tests.**

You asked for three; the seam has a few more edges worth a line each, so: older/newer/equal sequences (equal is the doc-write-then-row-update pair, which is load-bearing and was only covered end to end); a read recording nothing; `seal` rejecting a write dispatched *after* the delete but holding a lower number than the seal's fresh one; `LOCAL_WRITE` never recording *and* always admitted; and `clear` resetting the per-document records without restarting the counter.

That last one is the one I'd most have regretted leaving implicit — the counter surviving `clearAll` is what stops a stamp held by a request in flight across the clear from outranking one minted after it.

**3. `Math.max` in `seal` — dropped it rather than kept it as a guard.**

You're right that it's dead: every number in `applied` was minted by an earlier `next` or `seal`, so `++counter` already outranks all of them. I dropped the `max` instead of just fixing the comment, because the invariant that makes it dead ("the record moves forward because the counter does") is worth stating plainly, and a guard that cannot fire invites the next reader to look for the case where it does. The comment now states the invariant instead.

**Counts after the change:** 160 passed / 160 across `src/data-fetching` (was 153) — `writeGate` 7, `staleWrites` 58, `useDoctype` 24, `useCall` 18, `useDoc` 16, `useList` 16, `useNewDoc` 8, `docStore` 8, `useAction` 5. `vue-tsc --noEmit -p tsconfig.app.json` unchanged at the known 78-error baseline, none in any file this PR touches.

Not merging — leaving that to you.

PR #1052, opened 2026-08-11
fix(Dialog): raise z-index above sticky ListGroup headers (Closes #1051)

waterWang:
## What does this PR do?

Fixes #1051 — Sticky List Group headers render above open Dialogs.

The `ListGroup` sticky header uses `z-10` (from the `sticky top-0 z-10` Tailwind utility). When a `Dialog` is opened, `DialogPortal` teleports the overlay and scroll container to the end of `<body>`, but neither has an explicit z-index (defaults to `auto`). Because the sticky header is in a positive-z-index stacking context, it paints above the dialog backdrop and content.

## What changed

- Added `z-50` to the `DialogOverlay` class
- Added `z-50` to the dialog scroll container class

This ensures the dialog backdrop and content always render above sticky page elements (ListGroup headers, sticky table headers, etc.).

## Screenshots

Before: the sticky ListGroup header pokes through the open dialog.
After: the dialog fully covers the sticky header.

## Testing

- [x] Ran the existing Dialog Cypress tests
- [ ] Manual verification

No wallet address — pure contribution fix.

Coverage: 71.88% (±0.00% vs `main`)

PR #1064, opened 2026-08-12
fix(FloatingWindow): drop to z-40 and teleport to the host portal target

nextchamp-saqib:
A `FloatingWindow` sat at `z-index: 50`, the same layer as `Dialog`. At a tie the
painting order is document order, so a dialog the app renders before the window
loses and is unreachable.

Two changes:

**Layer.** The window drops to `z-40`: above ordinary page chrome, which tops out
at `z-10` (`ListGroup`'s sticky header, `PageHeader`), and below every dialog at
`z-50`. Popover surfaces are unaffected. Every one in this library carries
`z-[100]` on its reka content, which reka copies onto the teleported wrapper, so
the global `has-floating-window` style that used to lift them is no longer
needed and is removed.

**Portal target.** The window hardcoded `Teleport to="body"`. It now resolves the
host target through `usePortalTarget`, like the seven other native `Teleport`s in
the library. An overlay that resolves a host target would otherwise land before
the window and paint under it.

`usePortalTarget.spec.ts` scanned `src/` only, which is how this went unnoticed:
`experimental/` ships in the same package. The guard now covers both roots, and
it caught a second offender in `SpriteIcons/IconPicker.vue`.

Measured in Chrome, not asserted from the diff:

| case | before | after |
|---|---|---|
| sticky `z-10` group header over the window | header painted through the title bar | window on top |
| `Dialog` over the window | tie at `z-50`, decided by order | dialog on top |
| `DatePicker` popover inside the window | wrapper `z-100` | unchanged |

The layering spec now provides the host portal target. Without it both nodes land
in `<body>` and the assertion passes against the old hardcoded `body` target too.

Coverage: 71.92% (±0.00% vs `main`)

### DIFF
--- a/docs/content/docs/experimental.md
+++ b/docs/content/docs/experimental.md
@@ -105,6 +105,11 @@ composer-style windows. `v-model:mode` holds the state (`docked` | `floating` |
 scrollable body. `useFloatingWindow` is the headless half — pass it the panel
 and drag-handle refs to build your own chrome.
 
+A detached window sits at `z-index: 40`, above page chrome and below every
+dialog. It was `50` before, which tied with `Dialog`. If your app has chrome in
+that band, scope a rule to the `has-floating-window` class, set on `<body>`
+while a window is detached.
+
 ```ts
 import { FloatingWindow, useFloatingWindow } from 'frappe-ui/experimental'
 ```
--- a/experimental/FloatingWindow/FloatingWindow.cy.ts
+++ b/experimental/FloatingWindow/FloatingWindow.cy.ts
@@ -1,4 +1,5 @@
 import { defineComponent, h, ref } from 'vue'
+import Dialog from '#components/Dialog/Dialog.vue'
 import FloatingWindow from './FloatingWindow.vue'
 import type { WindowMode } from './types'
 
@@ -278,7 +279,9 @@ describe('FloatingWindow', () => {
         clientX: 800,
         clientY: 700,
       })
-      cy.get('body').trigger('pointercancel', { eventConstructor: 'PointerEvent' })
+      cy.get('body').trigger('pointercancel', {
+        eventConstructor: 'PointerEvent',
+      })
       // Move after cancel must be a no-op (listener torn down).
       cy.get('body').trigger('pointermove', {
         eventConstructor: 'PointerEvent',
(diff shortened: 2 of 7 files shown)


netchampfaris:
@nextchamp-saqib barista raised four things here and I checked each against the tree. The first one is real and blocks the argument this PR rests on.

**1. `Teleport to="body"` is hardcoded** (`experimental/FloatingWindow/FloatingWindow.vue:8`). Confirmed. The PR removes the `z-index: 50` on the grounds that body order already layers dialogs above the window. That only holds while both land in `<body>`. Overlays resolve their target through `usePortalTarget`, so a host that provides a target inside the app or a shadow root puts the dialog somewhere else and the window paints over it again. Two neighbours in `experimental/` already do it right: `TextEditor/components/ImageViewerModal.vue:201` and `TextEditor/extensions/image-group/ImageGroupUploadDialog.vue:306`.

The guard test does not catch it: `src/composables/usePortalTarget.spec.ts:174` sets `SRC = join(process.cwd(), 'src')`, so `experimental/` is never scanned. Worth widening while you are in here.

**2. `has-floating-window` has no reader left.** Confirmed. After this PR removes the style block, `git grep` finds only the setter (`useFloatingWindow.ts:24`) and the test cleanup (`FloatingWindow.cy.ts:68`). Drop both, or add a line saying it stays as a hook for app CSS.

**3 and 4** are fair nits: no test pins the new layering (a Cypress `document.elementFromPoint` check at the window centre with a dialog open would do it), and `docs/content/docs/experimental.md:98` should carry the stacking note from your PR description. An app with a `z-10` sticky header now paints over the window, which is new behaviour.

I did not push to your fork. Happy to if you would rather I take it.

netchampfaris:
Pushed `8f0ffb4d4`. Main finding fixed, two nits fixed, one declined on a wrong premise, and one of your suggestions replaced because it does not test what it looks like it tests.

### Portal target

Fixed. The window now resolves `usePortalTarget()` like every overlay. Measured in Chrome with a host target provided at `#host-portal`:

| | hardcoded `body` | after |
|---|---|---|
| window parent | `BODY` | `host-portal` |
| dialog parent | `host-portal` | `host-portal` |
| window before dialog in document order | **no** | yes |

So the ordering argument in the PR body only held without a host target, exactly as you said.

The guard now scans `experimental/` as well. That turned up a second offender, `experimental/SpriteIcons/IconPicker.vue`, whose `ComboboxPortal` had no `:to`. Fixed in the same commit. It also produced one false positive: `FloatingWindow.vue`'s own script comment names `<Teleport>` in prose and the regex read it as an unbound tag, so the scan drops `<script>` blocks before matching.

### The layering test

Your suggested assertion does not work, and it would have passed while broken. reka sets `pointer-events: none` outside a modal dialog, so `document.elementFromPoint` at the window centre misses the window whatever the paint order is. Measured with the hardcoded `body` target still in place, where the window *does* paint above:

```
elementsFromPoint(646, 416) -> [div.fixed.inset-0 (overlay), HTML]
```

No `.floating-window` in the stack at all. The test would have been green on the bug.

Replaced with a document-order assertion, which did discriminate: `false` with the hardcoded target, `true` after the fix. It also asserts neither element sets a z-index, since that is what makes document order decide.

### `has-floating-window`

Kept, with a doc comment saying it is the app's styling hook. Nothing in the library reads it, but it is the handle an app needs now that we set no z-index, so deleting it takes away the replacement at the same time as the thing it replaces.

### Docs

Declined as written. `docs/content/docs/experimental.md:98` is the Editor section; FloatingWindow has no docs entry anywhere in `docs/`. The stacking note went into the component's template comment instead, where a reader of experimental code will find it. A docs page for the component is worth its own PR.

Checks run locally: `usePortalTarget.spec.ts` 10/10 pass, `vue-tsc` clean for the touched files, prettier clean. The Cypress spec is not run — the binary does not start on this machine.

netchampfaris:
Both points hold. Fixed, and the branch is rebased onto `main` so it now has #1052's `z-50` Dialog.

**The layer.** You were right that removing it outright was wrong. Reproduced first, with a sticky `z-10` group header under a floating window:

```
elementsFromPoint(37, 25)
→ ["div.sticky.top-0[hdr]", "span.truncate.text-sm", ..., "div.floating-window"]
```

The header painted through the title bar, so the window read "Group 1" instead of its own title. At `z-40` the same point returns the title span first.

**The diagnosis.** Also right. The unreachable-dialog case is not this library's Dialog. At the old `z-50` the window tied with it and the tie went to document order, so a dialog the app renders *before* the window loses. That is what `z-40` fixes, and the PR body now says so instead of the wrong claim.

Checked before assuming `z-40` was safe below: every popover surface in this library carries `z-[100]` on its reka content, which reka copies onto the teleported wrapper (measured: a DatePicker opened inside the window gets `z-index: 100`). So the global `has-floating-window` rule that used to lift them stays removed.

**The test nit.** Correct — it passed against the old hardcoded `body` target. It now provides the host target through `global.provide` and asserts `win.parentElement === host`, plus the two z-index bounds instead of the document-order check.

Title and body updated; "stop setting z-index" is no longer what this does.

netchampfaris:
All three fixed. The IconPicker one was a real bug and it corrected a false claim I had made in the code comment.

**`IconPicker.vue:188`** — `z-10` was the only reka content in the library below 100, so my "every one carries `z-[100]`" had a counterexample in my own diff. Reproduced with the window at `z-40`:

```
wrapper z-index: 10
elementsFromPoint over the open menu
→ ["div.min-h-0.flex-1", "div.floating-window", "svg", "button.w-8.h-8", ...]
```

The window body sat above the dropdown, with the icon buttons buried under it. At `z-[100]` the menu is on top. The comment now names the exception instead of claiming it does not exist.

**The `> 15` floor** — right, and worse than stale: the scan finds 24 tags now, so deleting the whole `experimental` root would have passed the guard this PR just added it to. Floor raised to 20, comment carries the real counts.

**`MultiEmailInput.vue:64`** — right, and a good catch on why the scan misses it: the literal lives in `<script>`, which the new `SCRIPT_BLOCK` strip removes. `portalTo` is undefaulted now and resolved through `usePortalTarget`, like every consumer in `src/`.

**Docs** — added a sentence naming the `z-40` layer, the change from `50`, and the `has-floating-window` hook.

`usePortalTarget.spec.ts` 10/10 locally. Cypress still does not start on this machine, so those specs need CI.

Issue #881, opened 2026-07-30
Sweep: display and feedback


netchampfaris:
## Question

Bring the **display and feedback** family to bar for `1.0.0`.

Members: `Avatar`, `Badge`, `Alert`, `Toast`, `Skeleton`, `Spinner`, `LoadingIndicator.vue`, `LoadingText.vue`, `Icon`, `FeatherIcon.vue`, `Card.vue`, `ListItem.vue`, `Divider`, `Breadcrumbs`.

The biggest ticket by member count, but most members are small. Split it if it overruns a session.

Apply the at-bar checklist to each. Family-specific concerns:

- **Deprecated members to delete** under ADR-0008: `Card`, `ListItem`, `FeatherIcon`, and the `Toast` SFC (replaced by the imperative `toast(...)`, warns today). `spec/toast.md` is the Toast contract.
- `FeatherIcon` is the subtle one. `plan.md` originally wanted the export retained for back-compat so icon-name props (`Button.icon`, `Dialog.icon`, dropdown item icons) keep rendering feather names — but `deprecated-removals.md` lists it for deletion. Resolve that contradiction. Internal hardcoded usages migrate to `lucide-*` strings via the Tailwind plugin either way; use the `<span class="lucide-foo">` form, not `~icons/lucide/*` imports.
- `Spinner` / `LoadingIndicator` / `LoadingText` / `Skeleton` are four loading components. `P8` almost certainly applies to some of them.
- `Alert` — replace the type-specific `#icon` slot with `#prefix` (`P6`), add a uniform `icon?: string | Component` prop (`P11`), reconcile the `dismissible` / `theme` default drift between code and JSDoc, and add focus-visible on the dismiss button (`P12`).
- `Icon` versus `FeatherIcon` versus `lucide-*` strings versus a passed `Component` — settle the single recommended way to pass an icon, library-wide. This shows up in almost every other component's props.
- Issue #854: Icon Picker should allow custom search input and custom sections.

netchampfaris:
**Handoff from [#885](https://github.com/frappe/frappe-ui/issues/885): `CircularProgressBar` is yours.**

It is exported at `src/index.ts:138`, sitting inside the `// Charts` block between `AxisChart` and `DonutChart` — but it is not a chart, and it is not in this ticket's member list either. So it currently belongs to no sweep. #885 deletes everything else in that block; leave the export alone and just move it out of the block.

Two things come with it:

[...]

Issue #880, opened 2026-07-30
Sweep: command and shortcuts


netchampfaris:
## Question

Bring the **command and shortcuts** family to bar for `1.0.0`.

Members: `CommandPalette`, `KeyboardShortcut.vue`, `KeyboardShortcutsModal`, `SettingsDialog`, and the `useShortcut` composable.

Apply the at-bar checklist to each. Family-specific concerns:

- **None of these are in `v1-release/plan.md`'s core set.** They landed after it and have never had a v1 API audit.
- `CommandPalette` has a draft docs PR (#822) that has been open since early July — land it or close it, since a component cannot reach bar with no docs page.
- `SettingsDialog` overlaps the Dialog family. Is it a Dialog composition an app should own, or genuinely library surface? It has an open PR (#829, add close button). Coordinate with the Dialog ticket.
- `useShortcut` has a spec file (`useShortcut.spec.ts`) — confirm it is public API, documented, and its API matches how `CommandPalette` and `KeyboardShortcutsModal` consume it.
- `KeyboardShortcut.vue` is a bare `.vue` at the components root. Check whether it is exported and whether it should be.
- Keep-or-unexport call needed for each member — this family is a strong candidate for "an app should compose this itself".

netchampfaris:
From #870: `index.ts` may `export *` only from a curated barrel, never an implementation module — that habit is what published 31 unreviewed members across 6 lines. #928 converts all six to named exports mechanically, including the ones in your scope, which does **not** pre-empt your keep-or-remove calls. Please don't reintroduce a barrel star export when you sweep. The rule is recorded in `PHILOSOPHY.md` under P15.

netchampfaris:
[...]

Issue #878, opened 2026-07-30
Sweep: app shell


netchampfaris:
## Question

Bring the **app shell** family to bar for `1.0.0`.

Members: `DesktopShell`, `MobileShell`, `MobileNav`, `Rail`, `PageHeader`, `ScrollArea`, `Provider`.

Apply the at-bar checklist to each. Family-specific concerns:

- **None of these appear in `v1-release/plan.md`'s core set** — the whole family landed after that list was written. They have never been through a v1 API audit, and the freeze rule means they get one now or get un-exported.
- Naming: one namespace per family, variant as suffix (`PageHeaderMobile`, not `MobileHeader`). Check the family against that convention — `MobileShell` / `MobileNav` versus `DesktopShell` is worth a hard look.
- `Provider` — what does it provide, is it required, and is it public API or an implementation detail that leaked out?
- `ScrollArea` — keep-or-unexport call; is it a shell concern or a general primitive?
- Composability: expose slots rather than baked-in child components, so consumers only pay for the parts they use.
- These components define the shape of a whole Frappe app, so their APIs are the most user-visible thing in the library. Weight the "readable, idiomatic, simple in userland" standard heavily here.

`Sidebar` is a separate ticket; keep the vocabulary aligned with it.

netchampfaris:
Two inheritances from #870.

1. **The shells have no exported breakpoint.** `useIsMobile` and `useScreenSize` are un-exported (gameplan-only, 5 sites, and the library never used them itself). `DesktopShell` and `MobileShell` reference no breakpoint at all, so every app choosing between them now reinvents that number. Decide whether the shells should state one.

2. **Both shell components change** with the scroll rename in #928: `activeScrollContainer` → `shellScrollContainer`, `useScrollContainer` → `useShellScrolled`, and `registerScrollContainer`/`unregisterScrollContainer` become internal to `DesktopShell.vue` and `MobileShell.vue` (they were public via `export *` and no app ever called them).

netchampfaris:
[...]

Issue #877, opened 2026-07-30
Sweep: Tabs and TabButtons


netchampfaris:
## Question

Bring **Tabs** and **TabButtons** to bar for `1.0.0` — and resolve whether they stay two components.

Members: `Tabs`, `TabButtons`, and the internal `Pill`.

`v1-release/plan.md` names this as unresolved and must-fix-before-freeze: two overlapping public components covering the same concept. `P8` says a purely visual variant should be one component with a style axis, not two components.

Decide between:

- nest `TabButtons`' segmented rendering inside `Tabs` as a rendering mode
- merge into a single `Tabs` with a style axis
- keep both, with a written justification for why `P8` does not apply

Then apply the at-bar checklist to whatever survives.

Also:

- Three stale branches attack this from different angles — `refactor-tabs`, `tabs-rewrite`, `improved-tab-buttons`. Consolidate into one change set; do not land them separately.
- `Pill` is no longer exported (already done) but is retained internally by `TabButtons`, along with `PillSize`. Confirm that survives whatever restructuring happens.
- Whichever component survives owns both docs pages' content; the removed one needs a migration-guide entry.

netchampfaris:
**At bar ([#1024](https://github.com/frappe/frappe-ui/pull/1024), `feat(tabs)!`, merged 2026-08-10): the v0 monolith becomes a composed family and `TabButtons` aligns to it.** `spec/tabs.md` holds the contract and the decision log.

`Tabs` / `TabList` / `TabTrigger` / `TabPanel` replace the monolith, with a `tabs` shorthand kept for generated sets. **The model is the trigger `value`, never an index.** Route mode renders a trigger as a `RouterLink` and derives selection from the route, which deletes the `useActiveTabManager` and route-map helpers apps wrote by hand; a stale model selects the first visible trigger and emits, replacing the clamps apps wrote for conditional tabs. Four variants at two sizes, both orientations, RTL-aware, with `TabList` and `TabButtons` pixel-identical off a shared `src/components/shared/tabs/`. `iconRight` went per P11 — no story, doc or Figma pattern used it, and trailing content on a tab is a count or a badge, which `#suffix` carries.

**13 breaks, 6 of them silent** (index model, `type` → `variant`, `buttons` → `options`, `#prefix` → `#tab-prefix`, the `{ active: true }` fallback, `iconRight`).

[...]

PR #1068, opened 2026-08-12
refactor!: close four cross-family vocabulary mismatches before the freeze

### LINKED ISSUE #1054: Cross-family vocabulary pass before the freeze
Part of #864. Graduated from #1025 item 3 — it waited on #877, #881 and #944, all now closed.

## Question

Every family was reviewed in isolation. Do one pass **across** the final surface and fix real mismatches only.

What to check, against `CONTEXT.md`'s vocabulary and PHILOSOPHY.md (P6, P11):

- **`open`** — does every overlay family use it for the same thing (Dialog, Popover, Dropdown, ContextMenu, HoverCard, Tooltip, Sidebar, SettingsDialog)? Same prop name, same `v-model` shape, same emit.
- **Event names** — do similar events carry the same name across families (`update:modelValue` vs bespoke emits, `close` vs `update:open`, `change` vs `select`)?
- **Slot names** — the Tabs sweep landed on `#tab-prefix`/`#tab-suffix` while the input family uses `#prefix`/`#suffix` (#1044). Decide whether that divergence is justified by the shorthand-slot collision or is drift to close.
- **`variant` / `theme` / `size`** — same value vocabularies for the same visual axis.

**Fix only genuine mismatches.** A family that reads differently for a real reason keeps its wording; record why so the next pass does not relitigate it.

## Why this is a pre-tag ticket

#1025 says items 1 and 2 can run during the release-candidate week, but **this one cannot**: any rename it makes is a breaking change to a frozen export. Landing it after rc.1 would invalidate the soak, and after the tag it is a `2.0.0` change (map standing rule 2). It therefore blocks #1029.

Every rename this produces needs a migration-guide entry, and the codemod taught where one exists.

netchampfaris:
One pass across the whole exported surface for the four axes in #1054, then four renames. Every other divergence is recorded on the issue with the reason it stays, so the next pass does not relitigate it.

Full inventory and verdicts: https://github.com/frappe/frappe-ui/issues/1054

## The four renames

One commit each, so any one can be dropped.

### 1. Picker trigger slot props — `{ isOpen, togglePopover }` → `{ open, toggle }`

`DatePicker`, `DateRangePicker`, `DateTimePicker`, `TimePicker`.

`isOpen` is on CONTEXT.md's avoid list for public API, and `togglePopover` names the mechanism rather than the behavior (P1). #872 already made this exact rename on `Popover` — it is in the guide under "Popover / HoverCard / Tooltip → Slot props" — and #876 did not follow. The same three components already hand `#actions` a `close()`, so one component shipped two vocabularies for one concept.

**Blast radius:** custom `#trigger` / `#prefix` / `#suffix` on a date or time picker. Silent — a destructured `isOpen` becomes `undefined`, so a bound class stops applying with no error. **Four internal call sites**, all migrated here: two DatePicker stories, plus `experimental/Calendar/Calendar.vue` and its CustomHeader story, which drive the month/year picker from `#trigger`. Every app that already migrated `Popover` has done this rename once.

### 2. `SettingsDialog` — bare `v-model` → `v-model:open`

CONTEXT.md: `open` is "always bound via `v-model:open`"; bare `v-model` for overlay visibility is on the avoid list. Dialog's dual binding is a documented carve-out for Dialog alone, and P2 says new components do not propagate it. #880 kept the SettingsDialog family "as-is" on a keep-or-remove question and never argued the model name.

**Blast radius:** gameplan and suite, the two consumers named in #880. Silent — Vue accepts the unknown `modelValue` prop, so the dialog never opens. `v-model:tab` is unchanged.

### 3. `CommandPalette` — `v-model:searchQuery` → `v-model:query`

`Combobox` and `MultiSelect` bind the same controlled search text as `v-model:query`. One concept, one name (P1).

**Blast radius:** suite's sheets app, the consumer #880 kept the component for. Silent — the palette keeps its own query and the app's binding never updates. Documented next to #880's `show` → `open` row in the same guide section.

### 4. `Dialog` — `theme: 'yellow'` → `theme: 'amber'`

`DialogTheme` held the library's last `yellow`. Three pieces of evidence this is drift, not a considered choice:

- `Dialog.vue` rendered `yellow` with the **amber** tokens — `bg-surface-amber-2`, `text-ink-amber-5`. The value name disagreed with the token it resolved to.
- `utils/dialog.ts` carried `// Button doesn't have 'yellow'; fall back to default solid` — a special case that exists only because of the spelling.
- CONTEXT.md already lists `yellow` under the names **Alert** must avoid, and #881 moved Alert to `amber`.

**Blast radius:** `Dialog`'s `icon.theme` and the `theme` argument of `dialog.confirm` / `dialog.danger`. Silent for JavaScript (the tone map loses the key, the icon renders untinted), loud for TypeScript. `spec/dialog.md` is updated, and #873's `warning → yellow` mapping in the `icon.appearance` entry is corrected to `warning → amber` in the same commit.

## What was checked and left alone

- **`#tab-prefix` / `#tab-suffix` vs `#prefix` / `#suffix` (the ticket's slot question) — justified, no change.** `Tabs` is the only dual-mode component in the family; the unit prefix is what tells shorthand from composed at the call site, and per-tab slots are P6's repeated-unit case. `TabButtons` keeps plain `#prefix` / `#suffix`: it has one mode, and `Breadcrumbs` has the identical per-item shape with the same plain names. #1056's guide entry stands unchanged.
- **`setOpen` on the selection family** vs `toggle` on Popover — different signatures, documented in `spec/selection.md`.
- **`change` beside `update:modelValue`** on the pickers and Editor — nobody uses `change` *instead of* `update:modelValue`, so no concept carries two names.
- **`collapsed` on Sidebar**, **`expanded` on Tree**, **`dismiss` on Alert/SidebarCard**, **`TabsVariant`**, **`RailItem.variant`**, every `size` range — justified or settled by #872, #877, #878, #881. Reasons are on the issue.

Nine further findings that are not renames — including `Badge`'s undocumented deprecated `theme="orange"` alias, which ADR-0008 says cannot ship and which can only be discharged before the tag — are listed on the issue under "Handed on".

## Verification

- `npx vue-tsc --noEmit -p tsconfig.app.json` — **77 errors, unchanged from the `origin/main` baseline**, all in `vitepress/`, `src/mocks/handlers.ts`, `src/components/Menu/`, the editor extensions and `experimental/Charts`. Zero in any file this PR touches.
  - **Correction.** The first version of this PR said 78 and called it the baseline. It was 77 baseline plus one error this PR introduced — `experimental/Calendar/Calendar.vue(28,35): Property 'togglePopover' does not exist on type 'DatePickerTriggerSlotProps'`. It was missed by comparing per-file error *counts* instead of the error *list*. The Calendar call sites are fixed and the count is back to 77.
- Cypress, targeted — **170 / 170 passing** across 11 specs: DatePicker 23, DateRangePicker 21, DateTimePicker 15, TimePicker 20, SettingsDialog 7, SettingsPanel 8, CommandPalette 5, Dialog 22, `utils/dialog` 35, FrappeUIProvider 7, `experimental/Calendar` 7.
- Migration entries added for all four renames, each marked **silent**, in the existing before/after table style. Each rename also has a `## Unreleased` changelog entry naming the silent break and linking its guide section; the convention that requires both is recorded on #1054. Generated `*.api.md` regenerated for the components whose surface changed.
- No codemod change: the repo's only codemod (`tailwind/migrate-tokens-v2.js`) rewrites Tailwind classes and tokens and has no component-API mode, so the guide is the whole migration path — the same conclusion #877 reached for Tabs.
- Docs not built (`docs:build` OOMs on this box); the guide change is four table-and-prose blocks, checked for GFM table validity.

Part of #864
Closes #1054

Coverage: 71.85% (+0.02% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-1068/

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -47,7 +47,11 @@ The only two axes used to color components — there is intentionally **no** sem
 (`intent`/`severity`/`appearance`/`kind`/`status`):
 
 - **variant** — visual style: `solid | outline | subtle | ghost` (Button, Badge)
-- **theme** — color tone, by color name: `yellow | blue | red | green | amber | …` (Button, Badge, Alert, Dialog; the Alert/SidebarCard palette is `gray | blue | green | amber | red`)
+- **theme** — color tone, by color name: `gray | blue | green | amber | red | …` (Button, Badge, Alert, Dialog; the Alert/SidebarCard palette is `gray | blue | green | amber | red`)
+
+The warning tone is **`amber`** everywhere. No exported component accepts
+`theme="yellow"` — Dialog was the last one and moved to `amber` (#1054).
+_Avoid_: `yellow`
 
 A legacy `appearance` (`warning | info | danger | success`) maps to `theme` color names.
 
--- a/docs/components/recipes/DiscussionsDesktop.vue
+++ b/docs/components/recipes/DiscussionsDesktop.vue
@@ -741,7 +741,7 @@ const spaceActions = [
       </div>
     </DesktopShell>
 
-    <SettingsDialog v-model="showSettings" v-model:tab="settingsTab" size="5xl">
+    <SettingsDialog v-model:open="showSettings" v-model:tab="settingsTab" size="5xl">
       <SettingsSidebar>
         <SettingsNavGroup label="User settings">
           <SettingsNavItem value="profile">
--- a/docs/content/docs/changelog.md
+++ b/docs/content/docs/changelog.md
@@ -672,7 +672,7 @@ and `Rating` exports `RatingEmits`.
   independent of the auto-header).
 - Canonical slots `#default`, `#title`, `#actions` (scoped with
   `{ close, actions }`). The legacy `#body*` slots are removed — see below.
-- `icon.theme` (`yellow | blue | red | green`) replaces `icon.appearance`,
+- `icon.theme` (`amber | blue | red | green`) replaces `icon.appearance`,
   which is removed — see below.
 - Auto-header no longer renders an "Untitled" fallback.
 
@@ -709,7 +709,7 @@ aliased and nothing warns.
   dialog silently becomes dismissible. Use `dismissible`.
 - **Breaking, silent:** `icon.appearance` and `DialogIconAppearance` are gone;
   only `icon.theme` remains. An `appearance` key is dropped, so the icon
-  renders with no tone. Map `warning → yellow`, `info → blue`, `danger → red`,
+  renders with no tone. Map `warning → amber`, `info → blue`, `danger → red`,
   `success → green`.
 - **Breaking, silent:** the legacy `#body`, `#body-content`, `#body-main`,
   `#body-title` and `#body-header` slots are gone. Vue drops an unknown named
--- a/docs/content/docs/changelog.md
+++ b/docs/content/docs/changelog.md
@@ -197,6 +197,60 @@ Changed since `1.0.0-beta.41`, the first beta that shipped the family:
   `useChartTokens`, which re-resolves on a theme flip; `currentColorScheme` was
   the root `resolvedColorScheme` under another name.
 
+### DatePicker family — trigger slot props renamed to `open` / `toggle` (breaking, silent)
+
+`#trigger`, `#prefix` and `#suffix` on `DatePicker`, `DateRangePicker`,
+`DateTimePicker` and `TimePicker` now receive `{ open, toggle }` instead of
+`{ isOpen, togglePopover }` (#1054). `Popover`, `HoverCard`, `Dropdown`,
+`Select`, `Combobox` and `MultiSelect` all name the boolean `open`, and
+`Popover`'s trigger slot already names the flip `toggle`. `isOpen` is on
+`CONTEXT.md`'s avoid list for a public API, and `togglePopover` named the
+mechanism rather than the behavior (P1). The same
+components already gave `#actions` a `close()`, so one component shipped two
+vocabularies for one concept. `displayLabel` and `inputValue` are unchanged.
+
+**Silent break:** a destructured `isOpen` becomes `undefined`, so a class
+bound to it stops applying with no error; `togglePopover()` throws only if you
+call it. Grep for both names. See the
+[migration guide](/docs/migration#trigger-slot-props).
+
+### SettingsDialog — open state moves to `v-model:open` (breaking, silent)
+
+`v-model` → `v-model:open`, and `update:modelValue` → `update:open` (#1054).
+`CONTEXT.md` says the visibility of an overlay is `open`, "always bound via
+`v-model:open`", and lists bare `v-model` for visibility among the names to
+avoid. `Dialog`'s dual binding is a documented carve-out for `Dialog` alone,
+which P2 says new components do not inherit. `v-model:tab` is unchanged.
+
+**Silent break:** Vue accepts the unknown `modelValue` prop with no error, so
+the dialog never opens. See the
+[migration guide](/docs/migration#settingsdialog).
+
+### CommandPalette — `searchQuery` renamed to `query` (breaking, silent)
+
+`v-model:searchQuery` → `v-model:query`, and `update:searchQuery` →
+`update:query` (#1054). `Combobox` and `MultiSelect` already bind the
+controlled search text as `v-model:query`; one concept gets one name (P1).
+
+**Silent break:** the unknown `searchQuery` prop is accepted, so the palette
+holds its own query and your binding never updates. See the
+[migration guide](/docs/migration#commandpalette).
+
+### Dialog — `theme: 'yellow'` renamed to `'amber'` (breaking, silent)
+
+`DialogTheme` held the library's last `yellow` (#1054). `Alert`, `SidebarCard`,
+`Badge` and `Avatar` all spell the warning tone `amber`, and `Dialog` already
+rendered `yellow` with the amber tokens (`bg-surface-amber-2`,
+`text-ink-amber-5`), so the value name disagreed with the token it resolved to.
+Applies to `icon.theme` and to the `theme` argument of `dialog.confirm` /
+`dialog.danger`. Only the word changes, not the color.
+
+**Silent break** for JavaScript call sites: `yellow` is no longer a key in the
+tone maps, so the icon renders untinted and nothing throws. TypeScript call
+sites get a union error. This also corrects the `warning → yellow` mapping
+given in the `icon.appearance` entry below. See the
+[migration guide](/docs/migration#dialog).
+
 ### Charts (v1) family — moved to `frappe-ui/experimental` (breaking)
 
 The first chart family is not taken to bar at root for `1.0.0` (#942).
--- a/docs/content/docs/migration.md
+++ b/docs/content/docs/migration.md
@@ -2048,16 +2048,23 @@ usePageMeta(() => ({ title: pageTitle.value, emoji: '🌈' }))
 vocabulary. This is a **silent break**: Vue accepts the unknown `show` prop
 with no error, so the palette just never opens.
 
-| Before                  | After                    |
-| ------------------------ | ------------------------ |
-| `v-model:show="show"`    | `v-model:open="open"`    |
+`searchQuery` is renamed to `query`, the name `Combobox` and `MultiSelect`
+already use for the same controlled search text. Same silent break: the
+unknown `searchQuery` prop is accepted, so the palette holds its own query
+and yours never updates.
+
+| Before                          | After                    |
+| -------------------------------- | ------------------------ |
+| `v-model:show="show"`            | `v-model:open="open"`    |
+| `v-model:search-query="q"`       | `v-model:query="q"`      |
+| `@update:searchQuery="onSearch"` | `@update:query="onSearch"` |
 
 ```vue
 <!-- Before -->
-<CommandPalette v-model:show="show" :groups="groups" @select="onSelect" />
+<CommandPalette v-model:show="show" v-model:search-query="q" :groups="groups" @select="onSelect" />
 
 <!-- After -->
-<CommandPalette v-model:open="open" :groups="groups" @select="onSelect" />
+<CommandPalette v-model:open="open" v-model:query="q" :groups="groups" @select="onSelect" />
 ```
 
 `Mod+K` now opens the palette on its own (registered internally via
--- a/docs/content/docs/migration.md
+++ b/docs/content/docs/migration.md
@@ -144,6 +144,36 @@ throwing. TypeScript callers get a compile error instead. `#target` is the
 one slot case — content in a leftover `<template #target>` silently stops
 rendering.
 
+### Trigger slot props
+
+`#trigger`, `#prefix` and `#suffix` receive `{ open, toggle }` — the same two
+names `Popover`, `HoverCard`, `Dropdown`, `Select`, `Combobox` and
+`MultiSelect` hand out. `TimePicker`'s `#suffix` follows.
+
+| Before          | After    |
+| --------------- | -------- |
+| `isOpen`        | `open`   |
+| `togglePopover` | `toggle` |
+
+`displayLabel` and `inputValue` are unchanged, and `#actions` already used
+`close` — that stays too.
+
+This is a **silent break**: a destructured `isOpen` becomes `undefined`, so a
+class bound to it stops applying with no error, and `togglePopover()` throws
+`togglePopover is not a function` only if you call it.
+
+```vue
+<!-- Before -->
+<template #trigger="{ togglePopover, isOpen }">
+  <Button :class="isOpen && 'ring-2'" label="Pick a date" @click="togglePopover" />
+</template>
+
+<!-- After -->
+<template #trigger="{ toggle, open }">
+  <Button :class="open && 'ring-2'" label="Pick a date" @click="toggle" />
+</template>
+```
+
 Behavior changes that apply even if you don't touch your code:
 
 - `DateRangePicker` emits a `[from, to]` tuple. Update handlers that called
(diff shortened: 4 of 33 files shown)


netchampfaris:
Both points were right, and the blocker was worse than a miss — `vue-tsc` **had** flagged it and I read past it.

`experimental/Calendar/Calendar.vue(28,35): Property 'togglePopover' does not exist on type 'DatePickerTriggerSlotProps'` was in my own typecheck output. I compared the baseline and the post-change run as per-file error *histograms*, saw `experimental/Calendar/Calendar.vue 1` and assumed it was one of the known baseline errors. It was mine. The real baseline is **77**, not the 78 I reported. The PR body now carries that correction rather than the old number.

## Fixed

- `experimental/Calendar/Calendar.vue:28,34` and `experimental/Calendar/stories/CustomHeader.vue:113,119` — `#trigger="{ togglePopover }"` → `#trigger="{ toggle }"`.
- `spec/date-picker.md:329` and `:479` — both v1-contract lines now read `{ open, toggle, displayLabel, inputValue }`. Line 329 also still listed a `target` slot, which v1 renamed to `trigger`; corrected in the same line, since fixing half of a wrong sentence is not fixing it.
- `CONTEXT.md:50` — found by the re-grep. It led the `theme` value list with `yellow` and named `Dialog` as a user. Dialog was the only thing justifying the word, so after rename 4 the vocabulary doc was advertising a value no exported component accepts.

`experimental/Calendar/useEventBase.ts:115` — you were right, left alone. Both callers (`CalendarWeekDayEvent.vue:32`, `CalendarMonthEvent.vue:25`) pass `() => (isPopoverOpen = !isPopoverOpen)`, and `isPopoverOpen` is a local `ref(false)` in each (`:144` and `:97`). It never touches a picker slot.

## Re-grep, per rename

Scope each time: `src/`, `experimental/`, `docs/`, `vitepress/`, `spec/`, `skills/`, `tailwind/`, `CONTEXT.md`, `PHILOSOPHY.md`, all stories and specs. My original grep covered `src/` and `docs/` only, which is exactly the hole `experimental/` fell through.

**1. `togglePopover` / `isOpen` (picker slot props)**

| Where | Verdict |
|---|---|
| `experimental/Calendar/Calendar.vue:28,34` | **broken — fixed** |
| `experimental/Calendar/stories/CustomHeader.vue:113,119` | **broken — fixed** |
| `spec/date-picker.md:329,479` | **stale — fixed** |
| `src/components/DatePicker/stories/Examples.vue`, `stories/Range.vue` | migrated in the original commit |
| `experimental/Calendar/useEventBase.ts:115` | unrelated param, local closure — left |
| `experimental/SpriteIcons/IconPicker.vue`, `experimental/TextEditor/components/FontColor.vue`, `vitepress/components/Docs/CommandPalette.vue:31` | own local refs / watch params, not slot props — left |
| `TimePicker.vue:45`, `Popover.vue`, `Dialog.vue`, `BottomSheet.vue`, `Combobox.vue`, `MultiSelect.vue`, `ContextMenu.vue`, `DatePicker*.vue` internals | internal refs, which CONTEXT.md explicitly permits — left |
| `spec/popover.md`, `spec/imperative-api.md`, `migration.md` Popover / Autocomplete sections | historical v0 records of the *old* Popover API — left |
| `spec/date-picker.md:102` | a quoted "before" snippet under "Issues to fix before v1" — left |

**2. `SettingsDialog` bare `v-model`**

No live call site missed. Only `docs/components/recipes/DiscussionsDesktop.vue:744` (migrated in the original commit) and the migration guide's own before/after. `PHILOSOPHY.md`, `skills/frappe-ui/*` and `spec/adr/0010` name the component but never its binding.

**3. `searchQuery` → `query`**

No live call site missed. Every remaining hit is a local ref of the same name in an unrelated component — `KeyboardShortcutsModal.vue:94`, `docs/components/recipes/FilesMobile.vue`, `FilesDesktop.vue` — or a `data-cy="search-query"` attribute in `MultiSelect.cy.ts`. One cosmetic leftover fixed: `CommandPalette.cy.ts:69`'s test title still said "updates search-query while typing".

**4. `theme: 'yellow'` → `'amber'`**

| Where | Verdict |
|---|---|
| `CONTEXT.md:50` | **stale — fixed** |
| `docs/components/Home/Showcase/col3.vue:44` | passes `'yellow'` to a **`Badge`**, whose union has never had `yellow`. A pre-existing docs bug, not fallout from this PR — the tag renders untinted today on `main`. Reported on #1054, not fixed here |
| `spec/foundations.md:147`, `docs/components/foundations/PalettePage.vue`, `skills/frappe-ui/TOKENS.md`, `tailwind/colorPalette.js`, `tailwind/figma-tokens-to-theme.js`, `tailwind/migrate-tokens-v2.js` | the `yellow` **token ramp**, which still exists and is untouched — this rename is about the `theme` prop's value vocabulary, not the palette — left |
| editor color/highlight extensions and their tests | the editor's own color-name list, unrelated to component `theme` — left |
| `migration.md:945,959` (Alert), `spec/dialog.md` | correct before/after content — left |

## Re-verified

- `npx vue-tsc --noEmit -p tsconfig.app.json` — **77 errors, `experimental/Calendar/Calendar.vue` no longer among them**. Nothing in any file this PR touches.
- Cypress — **170 / 170 across 11 specs**: DatePicker 23, DateRangePicker 21, DateTimePicker 15, TimePicker 20, SettingsDialog 7, SettingsPanel 8, CommandPalette 5, Dialog 22, `utils/dialog` 35, FrappeUIProvider 7, **`experimental/Calendar` 7**.

One thing worth flagging for the freeze: `Calendar.cy.ts`'s 7 tests would **not** have caught this. None of them opens the header's month/year picker, so the only signal was the type error I misread. The break was silent in exactly the way the migration entry warns consumers about.

netchampfaris:
Both fixed, and the grep is now exhaustive rather than scoped.

- `docs/content/docs/changelog.md` — `icon.theme` union and the `warning →` mapping both say `amber` (26cac09).
- `spec/imperative-api.md:240` — slot props name the boolean `open` (f9ff0a0). Checked the surrounding headings before touching it: nearest are `183:## 2. The contract` and `205:### 2.1 Five verbs, and that's it`, so it is normative, not the historical record at `:134`.

### Re-grep, all 1601 tracked files

Method this time: `git grep -n -I -w <term> -- .` with no path filter, over `git ls-files`. Both previous misses came from a path filter (`src/` + `docs/` last round, `docs/content/docs/migration.md` only the round before).

**`togglePopover` — 0 live call sites.** 37 hits, all accounted for:

| Where | What it is |
| --- | --- |
| `src/components/TimePicker/TimePicker.vue:40,45,383` | internal function name, bound out as `toggle`. `CONTEXT.md` allows internal refs |
| `experimental/Calendar/useEventBase.ts:115,131` | unrelated parameter. Fed a local closure from `CalendarWeekDayEvent.vue:32` and `CalendarMonthEvent.vue:25`, both over a local `const isPopoverOpen = ref(false)`. You were right; left alone |
| `docs/content/docs/migration.md` ×12, `changelog.md:1143,1258` | the Popover and Autocomplete migration records. Historical by definition |
| `spec/popover.md` ×8, `spec/imperative-api.md:134` | v0 Popover record |
| `spec/date-picker.md:102` | quoted "before" snippet under "Issues to fix before v1" |
| `docs/content/docs/migration.md:156,162-168`, `changelog.md:204-213` | this PR's own before/after rows |

**`isOpen` — 1 hit needed a change, `spec/imperative-api.md:240`, now fixed.** 126 hits total. Every other one is an internal `ref`/`computed`/watch parameter (`Popover.vue:81`, `Dialog.vue:222`, `BottomSheet.vue:78`, the three date pickers, `TimePicker.vue:186`, `Combobox.vue:405`, `MultiSelect.vue:316`, `ContextMenu.vue:82`, `useAutofocusOnOpen.ts`, editor internals, `vitepress/components/Docs/CommandPalette.vue:31`) or a doc recording the old API (`spec/popover.md`, `migration.md` before-columns, `CONTEXT.md:29`'s avoid list, `.claude/commands/barista-review.md:46`'s avoid list). No public prop, emit or slot prop is left named `isOpen`.

**`searchQuery` / `search-query` — 0 missed sites.** 29 hits: this PR's rows in `migration.md`/`changelog.md`, the migrated `CommandPalette.cy.ts:26-32` and `stories/Example.vue` (local ref names, bound to `query`), plus unrelated local refs in `KeyboardShortcutsModal.vue:94`, `docs/components/recipes/FilesDesktop.vue:469`, `FilesMobile.vue:491`, and a `data-cy="search-query"` selector in `MultiSelect.cy.ts:144`.

**`yellow` — 0 remaining after the two changelog lines.** 313 hits, 238 of them the token-ramp JSON: the token ramp (`espresso-v2-design-tokens/`, `tailwind/colors.json`, `tailwind/generated/colors.json`, `tailwind/colorPalette.js`, `tailwind/figma-tokens-to-theme.js`, `tailwind/migrate-tokens-v2.js:361`, `skills/frappe-ui/TOKENS.md`), editor color names (`color-palette.ts`, `swatches.ts`, highlight/color extensions and their CSS), `Rating.vue`'s `text-yellow-*` classes, chart-ramp prose, `PalettePage.vue`, and this PR's own before/after rows. `yellow` stays a token; no exported component accepts it as a `theme`.

One pre-existing bug the sweep surfaced and I did not fix, because it is on `main` and outside this PR: `docs/components/Home/Showcase/col3.vue:44` passes `theme="yellow"` to a `Badge`, whose union never had `yellow`. Reported on #1054 with the other hand-ons.

### Changelog convention — answered on #1054

You were right that it wanted settling. Checked the history: `Alert` and `Tabs` have no changelog entry **at all**, not merely no rename entry, so their absence is a coverage gap, not a convention. The header rule is explicit, so I added the four entries. Evidence and reasoning on #1054.

### Verification

- `npx vue-tsc --noEmit -p tsconfig.app.json` — **77 errors**, the true baseline, unchanged. Docs-only commits.
- Cypress, 11 specs, **170 tests, 170 passing**: DatePicker 23, DateRangePicker 21, DateTimePicker 15, TimePicker 20, SettingsDialog 7, SettingsPanel 8, Dialog 22, CommandPalette 5, `utils/dialog` 35, FrappeUIProvider 7, `experimental/Calendar` 7. In one batch run CommandPalette, `utils/dialog` and FrappeUIProvider each reported "1 test, 1 failing" before any test ran; all three pass in isolation and on a clean re-run. Same batch-ordering flake as the earlier rounds, not a code failure.

netchampfaris:
Both addressed. Seven new tests, and the picker `toggle` now carries `Popover`'s signature rather than a second meaning.

### 1. Tests for the slot props that broke

`b2ddaa4`. One `it()` per picker, following `Popover.cy.ts:39` ("exposes reactive open state to the `#trigger` slot"):

| Spec | Test | What it exercises |
| --- | --- | --- |
| `DatePicker.cy.ts` | `exposes open and toggle to the #trigger slot` | mounts `trigger: ({ open, toggle }) => …`, clicks, asserts the popover opens and a class bound to `open` flips |
| `DatePicker.cy.ts` | `toggle sets the open state when passed a boolean` | `toggle(true)` on an open picker is a no-op, `toggle(false)` closes, a bare `toggle()` still flips |
| `DateRangePicker.cy.ts` | `exposes open and toggle to the #trigger slot` | same shape, `[from, to]` model |
| `DateTimePicker.cy.ts` | `exposes open and toggle to the #trigger slot` | same shape |
| `TimePicker.cy.ts` | `exposes open and toggle to the #suffix slot` | `TimePicker` has no `#trigger`; `#suffix` is where it hands out the two names |
| `TimePicker.cy.ts` | `toggle sets the open state when passed a boolean` | separate implementation from `PickerShell`, so it gets its own |
| `Calendar.cy.ts` | `opens the month picker from the default header` | the call site that actually broke: `Calendar.vue:28` binds `#trigger="{ toggle }"` |

**Verified red, not just green.** Reverting `PickerShell.vue`'s slot props to `{ isOpen, togglePopover }` fails both new `DatePicker` tests (23 passing / 2 failing). Reverting `Calendar.vue:28,34` to `togglePopover` fails the Calendar case (7 passing / 1 failing). Both reverts undone.

One shape note: a `#trigger` click is not auto-wired on the pickers the way reka wires `Popover`'s, so the slot owns its own handler. The boolean-setting tests call `toggle` directly instead of clicking, because a click outside the content also dismisses the popover and would mask the difference between setting and flipping.

### 2. `toggle` signature — widened, not documented as a divergence

`e18481b`. Took your first option. `PickerShell.vue`, `DatePicker/types.ts` and `TimePicker.vue` now declare `(flag?: boolean | Event) => void` with `Popover`'s semantics: a bare call flips, a boolean sets, a DOM event is ignored so `@click="toggle"` does not read the event as `true`.

Recording the divergence was the wrong call here. `setOpen` earns its entry in `spec/selection.md` because it is a *different name* for a different signature, so the call site tells you which contract you are in. `toggle` on a picker and `toggle` on a `Popover` are the same word, and a reader has no way to see which one they hold. This PR exists to make one word mean one thing.

Widening only, so nothing that worked before changes behavior: `togglePopover` never took an argument. `spec/date-picker.md` (both slot lines) and the migration guide say so, and `TimePicker.api.md` is regenerated.

### Counts

- Cypress, 12 specs, **197 tests, 197 passing**. The five in scope: DatePicker **25** (+2), DateRangePicker **22** (+1), DateTimePicker **16** (+1), TimePicker **22** (+2), `experimental/Calendar` **8** (+1). Plus Popover 20, SettingsDialog 7, SettingsPanel 8, CommandPalette 5, Dialog 22, `utils/dialog` 35, FrappeUIProvider 7.
- `npx vue-tsc --noEmit -p tsconfig.app.json` — **77 errors**, the baseline, unchanged. Zero in any file this PR touches.
- The batch-ordering flake showed up twice again (CommandPalette once, FrappeUIProvider once, each reporting "1 test, 1 failing" before any test ran). Both pass in isolation and on a clean re-run. Unchanged from the previous rounds.

Issue #1053, opened 2026-08-11
Docs site: retire the pages for removed components and rehome the parked families


netchampfaris:
Part of #864. Graduated from #1025 item 1 — the sweeps it waited on (#877, #881, #942, #944) are all closed, so the final component list is known.

## Question

The docs site still advertises the pre-sweep library. Decide the final information architecture and make the pages match the code.

Concrete drift found on `main` (f517c7d1b):

- `docs/content/docs/components/circularprogressbar.md` — the component was **removed** (#875, PR #958). The page should go.
- `docs/content/docs/components/listview.md`, `calendar.md`, `texteditor.md`, `codeeditor.md` — all four families were **parked in `frappe-ui/experimental`** (#985, #1020, #1007, #939), but their pages still sit under **Components** alongside the supported surface. `docs/content/docs/experimental/` currently holds exactly one page (`accordion.md`).
- `docs/content/docs/components/charts.md` — the v0 charts family was parked (#942); the v2 family has its own generated section from `src/charts`.

The sidebar is generated in `docs/.vitepress/config.ts` (`buildSidebar()`, `sourceRoots`) from folders that ship stories, so some of these pages may be orphaned rather than linked — check which, because an orphaned page is still reachable by URL and still ranks in search.

Beyond the per-page fixes, answer the question #1025 actually asks: **does the site's shape still make sense?** The `Legacy components` page (`components/legacy.md`) and an `Experimental` section now overlap — a removed component and a parked one are different things and a reader needs to be able to tell which is which. Decide whether parked families get pages under Experimental, redirects, or nothing.

Does **not** gate the tag: #1025 records that this can run during the release-candidate week, because it changes no code the test apps build against.

netchampfaris:
Worked in PR #1067. **This ticket's evidence was wrong, and the error was mine.**

The drift list — `circularprogressbar.md` plus four parked families still filed under Components — was gathered from a working tree that was sitting on a feature branch, not on `main`. On `main` those pages do not exist: #958 deleted `CircularProgressBar` outright, and `2f5cbab38` colocated component docs next to their source, so each parking PR carried the family's page along with its code. The right check was enumerating the routes `buildSidebar()` and the proxy generator actually produce, which is what #1067 did.

**What was actually wrong is the part this ticket called "the actual question"** — and it was worse than a naming overlap:

[...]

PR #1072, opened 2026-08-12
refactor(Badge)!: remove the deprecated theme="orange" alias, harden the theme lookup

### LINKED ISSUE #1069: Badge still ships a deprecated theme="orange" alias, which ADR-0008 forbids at 1.0.0
Part of #864. Found by the cross-family vocabulary pass (#1054), which correctly handed it on rather than folding it into a rename PR.

## Question

`Badge.vue:34` keeps `theme="orange"` as a **deprecated alias** for `amber`.

[ADR-0008](../blob/main/spec/adr/0008-no-deprecated-members-in-1-0-0.md) is unambiguous: nothing marked deprecated ships in `1.0.0`, because the tag freezes it until `2.0.0`. This one slipped the sweep — it is absent from `v1-release/deprecated-removals.md` and from the migration guide, so no census ever counted it.

## Work

1. **Census first** (the sweep's rule 5): grep every consuming app for `theme="orange"` / `:theme="'orange'"` on `Badge`, and for any indirect binding that can produce the string. The pins that decide how far each app has to travel are on the map.
2. Delete the alias.
3. Migration-guide entry, marked **silent** if an unknown theme renders untinted rather than throwing — verify which.
4. Append to the standing per-app upgrade issues rather than opening new ones (the shape settled on #869).
5. Add it to `v1-release/deprecated-removals.md` so the list matches reality.

**Blocks rc.1 (#1029).** ADR-0008 admits no "ships but deprecated" state at the tag, and removing it afterwards is a `2.0.0` change.

netchampfaris:
Two commits: the ADR-0008 removal, and a fix for the crash the removal exposed.

`Badge` kept `theme="orange"` as a deprecated alias for `amber`. ADR-0008 allows no "ships but deprecated" state at `1.0.0`, so it goes before the tag. `theme="amber"` renders exactly what `orange` used to.

The alias carried **no `@deprecated` tag** — only a line comment in `Badge.vue`. That is why no mechanical scan saw it, why it was never on `v1-release/deprecated-removals.md`, and why the sweep missed it. It reached the list late via the cross-family vocabulary pass (#1054). It is on the list now, with that reason recorded next to it.

## The removal exposed a crash, so this PR fixes that too

I mounted a Badge with `theme="orange"` after the removal rather than reasoning about it. It threw:

```
TypeError: Cannot read properties of undefined (reading 'subtle')
    at ComputedRefImpl.fn (src/components/Badge/Badge.vue:53:42)
```

`Badge` indexed a class map by theme and then indexed that result by variant. An unknown theme made the second lookup read a property of `undefined`, and Vue propagated the error up — the screenshot showed an empty mount area, so the badge vanished **and took the parent render with it**. With ~70 affected sites across 11 apps, that was ~70 blanked pages rather than ~70 wrong-coloured badges.

`resolvePropValue` (new, `src/utils/`) now backs all three lookup axes. An unsupported value falls back to the prop's default (`gray` / `subtle` / `md`) and warns once per offending value in dev:

```
[frappe-ui] Badge.theme="orange" is not a supported value — falling back to
"gray". Supported: gray, blue, green, amber, red, violet.
```

I probed `variant` and `size` before changing them: **neither ever threw.** They end their lookup chains, so they rendered untinted or unsized — quieter, but still wrong. They now warn too.

**The public type is not widened.** `orange` stays out of the union. TypeScript users still get a compile error; the runtime fallback is a net for JS call sites and bound values (`:theme="statusMap[s]"`) that TypeScript cannot see through. It is an upgrade cushion, not a supported way to pass a colour — the migration guide says so explicitly.

The warning follows the existing `warnUnsupportedIconString` idiom (DEV-only, deduped, names component/prop/value) rather than inventing a pattern. `warnDeprecated` was the wrong helper — its wording is deprecation-specific.

## Failure mode after this PR

| Call site | Behaviour |
| --- | --- |
| TypeScript | **Loud** — `vue-tsc` rejects the value at compile time |
| JavaScript / bound value | **Silent** — renders `gray`, one-time dev warning, nothing in prod |

## Census

49 direct `theme="orange"` Badge hits across 11 repos, plus 21 indirect bindings (ternaries, status maps, exported `BadgeTheme` unions).

| App | Direct | Indirect |
| --- | --- | --- |
| crm | 14 | 2 |
| lms | 11 | 2 |
| wiki | 7 | 0 |
| helpdesk | 4 | 1 |
| insights | 4 | 0 |
| central | 2 | 7 |
| suite | 2 | 3 |
| Letters | 2 | 0 |
| builder | 1 | 1 |
| press | 1 | 3 |
| slides | 1 | 0 |
| frappe (`ui/` package) | 0 | 2 |
| gameplan | 0 | 0 |

Because of the fix above, **these sites now render a grey badge rather than crashing a page.** They still need migrating — a grey badge where a coloured one belongs is a bug, and the dev warning is the only thing that will surface it.

Two sites reach more callers than they look like they do:

- `helpdesk/desk/src/components/PageInfo.vue:22` — `:theme="badge.theme ?? 'orange'"`. Every caller that omits `theme` lands on the fallback, so all of them go grey.
- `frappe/frappe` `ui/src/components/DataImport/dataImport.ts:4-13` — `getBadgeColor()` returns `"orange"` for three statuses. This is the first-party `@framework/ui` package.

Two apps already noticed the vocabulary mismatch and shimmed it. This is the pattern for apps that keep their own colour names:

- `central/dashboard/src/lib/roles.ts:82` — `/** Avatar themes use amber where badges use orange. */` plus a translation map.
- `pilot` `Updates.vue` — `theme: tone === 'orange' ? 'amber' : tone`.

### Census caveats — a lower bound, not exhaustive

- GitHub code search indexes default branches only.
- Punctuation-heavy queries (`:theme="'orange'"`, `theme: 'orange'`) return zero despite those forms demonstrably existing, so **the indirect column is a lower bound**.
- The rate limit truncated the press and lms sweeps.
- `builder`'s `GlobalDomains.vue`, `shop`'s `UiStatusBadge.vue` and `press`'s `site.js` are high-confidence but unconfirmed attributions.

### Out of scope, noted not acted on

`Button` and `Alert` accept `theme="orange"` at 5 sites across suite, Letters and toolbox. Neither documents an `orange` theme, so those are likely broken today already. Outside this ticket.

## Changes

- Deleted the alias computed and `'orange'` from the `BadgeProps` theme union.
- Added `src/utils/resolvePropValue.ts` and routed Badge's three lookups through it.
- Migrated internal call sites to `amber`: the docs dev-branch badge (2), the homepage showcase, the ContextMenu story, and the MultiSelect `TagsTrigger` story including its local type union and a `{ theme: 'orange' }` map entry.
- Kept the probe's finding as a real regression test: six specs covering fallback rendering for all three axes, the dev warning's contents, and its once-per-value dedup.
- Regenerated `Badge.api.md` via `yarn docs:gen`.
- Migration guide, changelog (removal + fix entries, deprecation-log row), and the `deprecated-removals.md` row.
- `spec/` needs no change: no Badge spec exists, and `orange` appears zero times under `spec/`.

## Verification

- `npx cypress run --component --spec src/components/Badge/Badge.cy.ts` — **17/17 pass** (11 existing + 6 new).
- ContextMenu **14/14**, MultiSelect **30/30**.
- `npx vue-tsc --noEmit -p tsconfig.app.json` — **77 errors**, matching the known baseline (vitepress/, Menu, src/mocks, editor extensions). **Zero in changed files.**
- Prettier clean on the new and changed source files.

Note for reviewers: ContextMenu and MultiSelect each reported a full-spec failure once when several agents shared this box, then passed on re-run. That is a dev-server port collision, not a flake in the specs.

## Expected conflict

`docs/content/docs/migration.md` is being rewritten concurrently by #1055, and `docs/content/docs/` pages by #1053. The Badge entry here is additive (one new `## Badge` section between `## Alert` and `## Sidebar`), so expect a mechanical conflict there and resolve by keeping both.

## Upgrade text for the per-app issues — NOT posted

I did not write to any other repo. Text for a human to append to the standing per-app upgrade issues:

> **`Badge theme="orange"` is removed** (frappe/frappe-ui#1069, part of frappe/frappe-ui#864).
>
> Replace with `theme="amber"`. It renders identically — `orange` was only ever an alias for `amber`.
>
> **How you find out you missed a site depends on your call sites.** TypeScript rejects the value at compile time. JavaScript and bound values render a **grey** badge and log a one-time dev-mode warning; production logs nothing. So a missed site is a wrong-coloured badge that only the dev console reports.
>
> Grep for all three shapes — the last two are the ones that bite:
> - `theme="orange"` — the literal attribute.
> - `:theme="... ? 'orange' : ..."` — ternaries and status-to-theme maps. `vue-tsc` will not always catch these.
> - `?? 'orange'` / `|| 'orange'` — an orange **default** affects every caller that omits a theme, not just the ones asking for orange.
>
> Watch the dev console for `[frappe-ui] Badge.theme="orange" is not a supported value` while you soak — that is the reliable signal for the bound cases.
>
> If your app keeps its own colour vocabulary and cannot rename `orange` at the source, translate at the boundary: `const badgeTheme = tone === 'orange' ? 'amber' : tone`. `central` and `pilot` already do this.

Part of #864
Closes #1069

Docs preview: https://ui.frappe.io/pr-preview/pr-1072/

Coverage: 71.88% (+0.05% vs `main`)

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -49,6 +49,9 @@ The only two axes used to color components — there is intentionally **no** sem
 - **variant** — visual style: `solid | outline | subtle | ghost` (Button, Badge)
 - **theme** — color tone, by color name: `yellow | blue | red | green | amber | …` (Button, Badge, Alert, Dialog; the Alert/SidebarCard palette is `gray | blue | green | amber | red`)
 
+Badge themes are `gray | blue | green | amber | red | violet`.
+_Avoid_ (Badge): `orange` — an alias for `amber`, removed in `1.0.0` (ADR-0008)
+
 A legacy `appearance` (`warning | info | danger | success`) maps to `theme` color names.
 
 **theme** means color tone and nothing else. It is the library's most-used
--- a/docs/.vitepress/theme/Layout.vue
+++ b/docs/.vitepress/theme/Layout.vue
@@ -61,7 +61,7 @@ onBeforeUnmount(() =>
           <Badge
             v-if="devBranch"
             :title="`git branch: ${devBranch}`"
-            theme="orange"
+            theme="amber"
             variant="outline"
             class="hidden md:flex"
           >
@@ -90,7 +90,7 @@ onBeforeUnmount(() =>
           <Badge
             v-if="devBranch"
             :title="`git branch: ${devBranch}`"
-            theme="orange"
+            theme="amber"
             variant="outline"
             class="hidden md:flex"
           >
--- a/docs/components/Home/Showcase/col2.vue
+++ b/docs/components/Home/Showcase/col2.vue
@@ -149,7 +149,7 @@ const resetState = () => {
         <label class="w-full">Status</label>
 
         <Badge theme="blue" class="rounded-1"> Stable</Badge>
-        <Badge theme="orange" class="rounded-1"> Moderate</Badge>
+        <Badge theme="amber" class="rounded-1"> Moderate</Badge>
       </div>
     </div>
 
--- a/docs/content/docs/changelog.md
+++ b/docs/content/docs/changelog.md
@@ -9,19 +9,37 @@ one-time dev-mode warning (unless noted). Removal is post-v1.
 
 ## Unreleased
 
-### Badge — `theme="orange"` removed (breaking, loud)
+### Badge — `theme="orange"` removed (breaking; loud in TS, silent in JS)
 
 `orange` was a deprecated alias that resolved to `amber`, so `theme="amber"`
 renders exactly what `theme="orange"` used to. ADR-0008 keeps nothing
 deprecated in `1.0.0`, and the alias was never on the removal list, so no
 earlier census counted it (found by #1054).
 
-**Loud break:** `Badge` indexes a class map by theme, so an unknown value
-throws `TypeError: Cannot read properties of undefined (reading 'subtle')`
-while rendering. The badge does not appear, and the error takes the parent
-render with it. TypeScript call sites fail earlier, at `vue-tsc`. Check bound
-themes as well as literal attributes — a status-to-theme map that yields
-`'orange'` throws the same way and no grep for `theme="orange"` finds it.
+TypeScript call sites fail at `vue-tsc`, because the `theme` union no longer
+accepts the string. JavaScript call sites and bound values render the default
+`gray` theme and log a one-time dev-mode warning naming the component, prop and
+value. Check bound themes as well as literal attributes — a status-to-theme map
+that yields `'orange'`, or an `?? 'orange'` default, goes grey and no grep for
+`theme="orange"` finds it.
+
+### Badge — an unsupported `theme`, `variant` or `size` no longer crashes (fix)
+
+`Badge` indexed a class map by theme and then indexed that result by variant.
+A theme outside the union made the second lookup read a property of
+`undefined`, throwing `TypeError: Cannot read properties of undefined (reading
+'subtle')` mid-render — the badge vanished and the error took the parent render
+with it, so one stale colour name in a status map could blank a page.
+
+All three axes now fall back to the prop's default (`gray` / `subtle` / `md`)
+and warn once per offending value in dev. `variant` and `size` never threw —
+they ended their lookup chains and rendered untinted or unsized — but they were
+silently wrong, and now report themselves too.
+
+The public types are unchanged: `theme` still accepts only the six supported
+values, so TypeScript keeps rejecting anything else at compile time. The
+fallback is a runtime net for JS call sites and bound values, not a widening of
+the API.
 
 ### Editor — images and embeds resize from a bottom-right corner handle
 
@@ -1790,4 +1808,4 @@ names.
 | `useFileUpload` / `FileUploadHandler` unset privacy | explicit `private` / `is_private` | **Default changed** — silent; now resolves to private |
 | `fileToBase64`, `formatBytes`, `getMaxFileSize`, `fileSizeLimitMessage` | none (internal only) | **Removed** — import fails |
 | `frappe-ui/charts` `ColorScheme` type | root `ResolvedColorScheme` (re-exported from `frappe-ui/charts`) | **Removed** — loud; type import fails |
-| `Badge theme="orange"`             | `theme="amber"`                      | **Removed in 1.0.0** (ADR-0008) — loud; throws while rendering |
+| `Badge theme="orange"`             | `theme="amber"`                      | **Removed in 1.0.0** (ADR-0008) — loud in TS (compile error); silent in JS (renders gray, dev-only warning) |
--- a/docs/content/docs/changelog.md
+++ b/docs/content/docs/changelog.md
@@ -9,6 +9,20 @@ one-time dev-mode warning (unless noted). Removal is post-v1.
 
 ## Unreleased
 
+### Badge — `theme="orange"` removed (breaking, loud)
+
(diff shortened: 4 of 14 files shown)


netchampfaris:
Thanks — all four fixed in a720ed4d. Summary per nit, plus one correction to my own claim.

**1. `CONTEXT.md` `_Avoid_` entry.** Added next to the theme vocabulary, matching Alert's shape at `:88`:

```
Badge themes are `gray | blue | green | amber | red | violet`.
_Avoid_ (Badge): `orange` — an alias for `amber`, removed in `1.0.0` (ADR-0008)
```

**2. `deprecated-removals.md` counts and scope.** The arithmetic was wrong before my row too — "twelve of the thirteen … two rows sit outside" already totalled fourteen against a thirteen-row table. The table now has fourteen data rows with three outside the block, so it reads **eleven**.

On scope, I **widened the sentence rather than dropping the row**. A prop value is invisible to any scan of `src/index.ts`, and being absent from this list is the whole reason the sweep missed this alias — moving it to the migration guide alone would rebuild the exact gap #1069 exists to close. The sentence now covers deprecated prop values outright, not just ones that pull an export with them.

**3. `TagsTrigger.vue`.** Now `theme: BadgeProps['theme']`.

**Correction to your framing, and to my own first attempt at the comment:** this does *not* fail type-check today. `tsconfig.app.json` excludes `src/**/stories/**/*`, and the root config is solution-style over app/node/cypress, so no project includes stories. I verified both halves rather than assuming — with a temporary config that pulls the story into the program, the derived union is clean, and reintroducing `'orange'` produces:

```
TagsTrigger.vue(21,31): error TS2322: Type '"orange"' is not assignable to
type '"gray" | "red" | "blue" | "green" | "amber" | "violet" | undefined'.
```

So the derivation is sound and it fails in the editor, but it will not fail CI until stories join the type-check program. The comment in the file states that limit instead of implying a guard that runs. Removing the duplicated union is still the right fix — it just is not enforcement yet.

**4. `resolvePropValue` direct tests.** New `src/utils/resolvePropValue.test.ts`, 10 tests. Your catch was right: the Cypress spec only proved once-per-render.

I verified the dedup test actually pins the key rather than trusting that it does. Mutating `:46` to drop `=${value}` fails **exactly one** test and no others:

```
× resolvePropValue > warns once per distinct value, not once per component and prop
Tests  1 failed | 9 passed (10)
```

Restored, 10/10 green. Also covered: the `PROD` early-return, that `PROD` does not burn the dedup entry (a value first seen in prod must still warn in dev), and that inherited `Object.prototype` keys like `toString` fall back instead of returning a function where a class name belongs.

**Not acted on:** your closing note that the repo never runs `type-check` in CI, which makes the "loud in TypeScript" half of the migration story unverified here. It is correct and it matters — and nit 3 above is a second instance of the same gap — but it is repo-wide, not this PR's. Faris is filing it separately.

### Verification

| Check | Result |
| --- | --- |
| `Badge.cy.ts` | 17/17 pass |
| `resolvePropValue.test.ts` | 10/10 pass |
| `ContextMenu.cy.ts` | 14/14 pass |
| `MultiSelect.cy.ts` | 30/30 pass |
| `vue-tsc -p tsconfig.app.json` | 77 errors, matching baseline; zero in changed files |
| Prettier | clean on all new and changed files |

Note: `ContextMenu.cy.ts` reported 14/14 *failing* on one run, then 14/14 passing on re-run with no change between. `MultiSelect` did the same earlier. It is a dev-server port collision from parallel agents on this box, not a spec flake — worth knowing if CI shows one red run.

PR #1083, opened 2026-08-13
refactor(CommandPalette)!: rebuild as a seven-part family in experimental

netchampfaris:
`CommandPalette` is rebuilt as a family of composable parts and moves to
`frappe-ui/experimental`. The old one had one shape and no filtering, so four
apps forked it instead of using it. The parts fit all four.

It stays in `experimental` until gameplan, helpdesk and this docs site run on
it. No stability promise there: it can change shape or leave in any release.

## The seven parts

`CommandPalette` owns the dialog, the query, the keyboard and the filter. The
app writes the rows.

| Part | What it is |
| --- | --- |
| `CommandPalette` | the root |
| `CommandPaletteInput` | the search field |
| `CommandPaletteList` | the list, and the only part that scrolls |
| `CommandPaletteGroup` | a heading and its rows, hidden whole when the filter empties it |
| `CommandPaletteItem` | a row |
| `CommandPaletteEmpty` | the no-matches message |
| `CommandPaletteFooter` | the hint bar |

Groups and items go inside `CommandPaletteList`. The field, the empty state and
the footer are its siblings, because a list may own rows and groups and nothing
else.

## Every break against v0

- `import { CommandPalette } from 'frappe-ui'` no longer resolves. Import from
  `frappe-ui/experimental`.
- The `groups` prop is gone. Groups and rows are markup, so a row renders
  whatever it needs without a `component` escape hatch.
- `@select` carries two arguments, the value and the click that picked it.
  `event.preventDefault()` keeps the palette open.
- `Mod+K` moves to the caller. The old palette registered it and skipped it
  whenever a rich-text editor had focus, hardcoding the editor into the palette.
- Filtering is on by default. The old palette rendered `groups` as given. Set
  `:filterable="false"` for server search. That is the word `Combobox` and
  `MultiSelect` already use (ADR-0009).
- On `1.0.0-beta` or older, two more renames land in the same move: `show` to
  `open`, and `searchQuery` to `query`. Both are silent, because Vue accepts the
  unknown prop and the palette simply never opens or never updates. Suite's
  `SheetEditor` still binds the old names today.

## What a caller has to do

1. Import the parts from `frappe-ui/experimental`.
2. Register the shortcut. `allowInInput` defaults to `false`, so without it
   `Mod+K` dies the moment a field has focus:
   ```js
   useKeyboardShortcut({
     combo: 'Mod+K',
     allowInInput: true,
     handler: () => (open.value = true),
   })
   ```
3. Write the rows as markup, with groups and items inside `CommandPaletteList`.
4. Handle `@select` with `(value, event)`. Do not close the palette yourself: it
   already does, and closing again makes `preventDefault()` do nothing.
5. For server search, set `:filterable="false"` and refetch on `update:query`.

Before and after for each break is in the migration guide.

## Also

- Every part stamps `data-slot`. A row stamps `data-state` and `data-disabled`
  (P10) and hands `active` and `disabled` to its slots (P7).
- Built on reka-ui's Listbox. That was the last `@headlessui/vue` import in
  `src/`, so the dependency leaves `package.json` and the readme names Reka UI.
- A row filters on the text of its default slot. `#prefix` and `#suffix` are
  left out, so a trailing shortcut hint never becomes searchable.
- Docs page, three stories (client filter, debounced server search, link rows),
  API tables, migration section, changelog, legacy and experimental pages.

## Review

Agents ran four local review rounds against `.claude/commands/barista-review.md`
and fixed what they found. The ones that changed behavior:

- An item measured its label off the DOM, but that DOM only mounted once the
  filter passed. A palette opened with a query already set showed nothing, and
  never recovered until the query cleared. Two regression tests cover it.
- The field and the footer rendered inside `role="listbox"`, which may own only
  rows and groups. `CommandPaletteList` exists to fix that. The measured
  `scroll-padding` machinery it replaced is deleted, not patched.
- `aria-activedescendant` pointed at a row the filter had removed, so the empty
  state announced no results while the field still claimed a highlighted row.
  axe graded it critical.
- The listbox had no accessible name. It takes the dialog's `title`.
- `data-state="selected"` was unreachable. reka records a value only when the
  select event is not default-prevented, and `preventDefault()` is how a caller
  keeps the palette open. The `selected` slot prop went with it.

## Verified

- `npx vitest run`: 92 files, 1529 tests pass.
- `yarn build`: clean.
- `yarn docs:check`: committed API tables match the source.
- `vue-tsc`: 78 errors, the same 78 `origin/main` gives, none in
  `experimental/CommandPalette/`.
- axe on the docs page, populated, empty and narrowed-to-one-row: no violations
  except a library-wide `color-contrast` on `text-ink-gray-5` and
  `text-ink-gray-6` in dark mode, filed separately.
- Keyboard geometry measured in Chrome at two viewports, 40 rows, four
  positions, plus a footer grown after mount.

Not verified: `CommandPalette.cy.ts` has never executed. Cypress 15.9.0 cannot
start on macOS arm64 here, so CI runs it first. Screen-reader behavior is
verified as markup, not with a screen reader.

Docs preview: https://ui.frappe.io/pr-preview/pr-1083/

Coverage: 72.35% (-0.03% vs `main`)


netchampfaris:
## Local barista review, round 3

The bot review above ran once when this PR opened and never re-ran, so its 3/5
scores the first commit. Four review rounds ran locally instead, against
`.claude/commands/barista-review.md`. This is round 3, the last one in barista
format. Rounds 4 and 5 fixed what it found; that list is at the bottom.

---

**Concerns (3/5)** — the structural change is right and the layout claim was
proved in a real browser. One silent failure mode, plus nits.

### Findings

**Concerns**

1. `CommandPaletteItem.vue:75-79` (same for `CommandPaletteGroup.vue:40-44`) —
   no DEV warning when the caller forgets `CommandPaletteList`. Run it: the
   palette renders, there is no `role="listbox"` anywhere, arrow keys do
   nothing, `data-state` never appears, `aria-activedescendant` stays null, and
   nothing is logged. This round made the list mandatory, so the mistake is new.
   Every other misuse in this family warns.

**Nits**

2. `CommandPaletteEmpty.vue:2-8` — the message is outside the listbox now, so it
   is neither an option nor a live region. reka does not clear
   `aria-activedescendant` when the filter empties the list, so it still points
   at a removed row. A screen-reader user who types a query that matches nothing
   hears nothing.
3. `CommandPaletteItem.vue:93-102` — the "draws no text" warning false-positives
   on an item whose slot text fills one tick after mount.
4. `CommandPaletteEmpty.vue:7` — with no items and an empty query the default
   renders `No results for ""`. `empty` is true whenever nothing is visible, not
   only when the filter hid something, so two comments describe something the
   code does not compute.
5. Stale part counts in three code comments: `experimental.ts:35`,
   `src/index.ts:108`, `tailwind/content.js:42`.
6. `changelog.md:389-397` — the `searchQuery` to `query` entry survives next to
   the entry that removes the component from the root export.
7. `experimental.md:33` contradicts its own legend at `:21-25`.
8. `CommandPalette.md:3` says "opened with `Mod+K`", `:50-51` says the palette
   registers no shortcut.
9. `stories/ServerSearch.vue:59` — `@select="open = false"` is redundant and
   makes a caller's `preventDefault()` useless in the server-search reference.
10. Every palette mount logs reka's missing-`Description` warning. Origin is
    `Dialog.vue`, not this PR.

### The four judgments asked for

**ARIA roles are valid now.** Exactly one `role="listbox"`, on
`data-slot="command-palette-list"`; its only children are `role="group"`; their
children are `role="option"`. Field, empty state and footer are outside it.

**`CommandPaletteEmpty` as a sibling is correct.** It stops being an invalid
listbox child, and it stops being announced through the option collection.
Nothing replaces that announcement (nit 2).

**The flex layout holds. Proven, not reasoned.** Driven in real Chrome at
1000x700, 40 rows. Root 420px, field 41.1px `shrink-0`, list `min-height: 0`
with `scrollHeight 1330 / clientHeight 347`, footer 31.9px `shrink-0`.

| step | active row | inside list | clear of field and footer |
| --- | --- | --- | --- |
| 20 down | Row 21 | true | true |
| 39 down | Row 40 | true | true |
| 8 back up | Row 32 | true | true |
| back to top | Row 2 | true | true |

Footer grown to 83.9px after mount: the list gave up the space, 347 to 295px,
and Row 22 stayed clear. reka scrolls with `block: 'nearest'`, and chrome
outside the scroller cannot cover a row, so deleting the measured
`scroll-padding` loses nothing.

**Seven parts is right.** Only `ListboxContent` renders `role="listbox"`;
`ListboxRoot` renders no role, so the list part is not optional in reka's model.
It matches `List` + `ListRows` and `Tabs` + `TabList`. The six-part alternative,
the root rendering the content with `#footer` and `#empty` as named slots, would
have blocked what `ServerSearch.vue` does: swap the whole list for a loading
block.

### The `selected` removal

Checked against reka-ui 2.9.9. `ListboxItem.vue:52-58` is
`emits('select', ev); if (ev?.defaultPrevented) return; rootContext.onValueChange(...)`,
so `preventDefault` blocks the record and skips `changeHighlight`. The other
writers into `modelValue` are Cmd+A select-all and shift-range, both
`multiple`-only, and the palette never sets `multiple`. No reachable path, so
the removal is right, and P7 is better served by dropping a permanently-`false`
slot prop than by handing it out.

### Checked and cleared

92 files, 1525 tests, exit 0. `vue-tsc`: 78 errors, none in
`experimental/CommandPalette/**`. Prettier drift on 7 `.md` files also drifts on
`origin/main`. No source still imports the removed root `CommandPalette`.
`@headlessui/vue` removal complete. API tables document all seven parts.
Migration samples work against the real API. Stories use `lucide-*` classes and
no real emails.

---

## What the two rounds after this review changed

- Finding 1 fixed: `CommandPaletteList` provides a context, and the item and the
  group warn in DEV when it is missing.
- Nits 2 to 9 all fixed. The empty state got `role="status"`; the "draws no
  text" warning now waits for a query, because a tick-count re-check is a race;
  the default message branches on the query.
- An axe run then found two more, both fixed: `aria-activedescendant` pointing
  at a removed row (**critical**), and the listbox having no accessible name.
  It takes the dialog's `title`.
- CI ran `CommandPalette.cy.ts` for the first time and failed it. The hover test
  triggered `mouseover`; reka binds `onPointermove` and nothing binds
  `mouseover`, so the highlight never moved and the test never tested its name.
  Fixed, and the event name is now proved under vitest, which can actually run
  here. The other four tests were audited for the same class of dead call.
- Nit 10 is Dialog's, filed separately.

Final: 8/8 CI, Greptile 5/5 at `0c585387a`.

Agents ran the reviews and the fixes.

PR #1084, opened 2026-08-13
fix: remove Google Docs/Drive from iframe allowlist to prevent silent link loss (Fixes #1082)

### LINKED ISSUE #1082: Editor: pasted Google Drive/Docs links silently become iframe embeds that server-side sanitization strips
## Context

Reported by a customer via support ticket [#75814](https://support.frappe.io/helpdesk/tickets/75814) (Frappe Helpdesk v1.28.1, frappe-ui `1.0.0-beta.24`). Pasting a Google Drive share link into the ticket reply / new ticket editor and clicking Send loses the link entirely — the saved reply contains no trace of it, with no error shown. The customer flagged it as urgent; inserting the URL via the link toolbar button works as a workaround, but the paste path keeps silently eating links.

## Root cause

1. `RichTextKit` ships the `Iframe` extension by default (`src/molecules/editor/kits.ts`).
2. The iframe paste plugin's bare-URL branch (`src/molecules/editor/extensions/iframe/iframe-paste-handler.ts`, step 3) validates pasted plain text against `IFRAME_ALLOWLIST`, which includes `drive.google.com` and `docs.google.com`. A pasted Drive link is silently replaced with an `<iframe>` embed node — no confirmation and no way to paste it as a plain link. (The legacy handler's `window.confirm` prompt was removed in the rewrite, and its URL regex never matched Drive links, so this is a behavior change from the v0 TextEditor.)
3. Frappe sanitizes HTML fields on save via `frappe.utils.html_utils.sanitize_html`, and `acceptable_elements` does not include `iframe`. The server strips the embed, so the submitted content loses the link silently.

Editor round-trip is fine; the data loss happens at the consumer's backend. But every Frappe-backed consumer of `RichTextKit` (Helpdesk, CRM, …) inherits this: paste an allowlisted URL (Google Drive/Docs, YouTube, Vimeo, Figma, Notion, …) → content silently lost on save. Non-allowlisted URLs paste as plain text/links and survive, which is why only these hosts are affected.

## Steps to reproduce

1. Open a `RichTextKit`-based editor backed by a Frappe HTML field (e.g. Helpdesk ticket reply).
2. Paste `https://drive.google.com/file/d/xxxxxxx/view` as plain text.
3. The URL turns into an iframe embed. Submit.
4. Reopen the saved content — the link (and embed) is gone.

## Expected behavior

Pasting a URL must never silently destroy content. The link should survive the round-trip exactly as entered.

## Suggested direction

- Paste bare URLs as plain links by default and make embed conversion opt-in — e.g. a "convert to embed" affordance on the pasted link or a paste menu (Notion/Linear pattern). This keeps the embed feature while removing the silent destructive default.
- Alternatively (or additionally), keep the iframe insert dialog but drop the bare-URL branch of the paste handler; explicit embed snippets (`<iframe …>`) pasted deliberately can keep converting.

The server-side stripping is Frappe's long-standing sanitizer policy, so the fix belongs on the editor side: don't auto-create nodes that consumers' backends are known to discard.

waterWang:
## Problem

Pasting a Google Drive/Docs share link into the editor and clicking Send loses the link entirely — the saved reply contains no trace of it, with no error shown. Inserting the URL via the link toolbar button works as a workaround.

## Root cause

`RichTextKit` ships the `Iframe` extension by default. The iframe paste plugin's bare-URL branch validates pasted plain text against `IFRAME_ALLOWLIST`, which includes `docs.google.com` and `drive.google.com`. A pasted Drive link is silently replaced with an `<iframe>` embed node — no confirmation and no way to paste it as a plain link. When the content is saved, server-side sanitization strips the iframe, leaving nothing.

## Fix

Remove `docs.google.com` and `drive.google.com` from:
- `IFRAME_ALLOWLIST` in `iframe-allowlist.ts` — the paste handler and dialog won't auto-convert these URLs to iframe embeds
- `PLATFORM_CONFIGS` in `iframe-embed-utils.ts` — the corresponding platform entry is no longer needed

Users can still paste Google Drive links as plain text links (via the link toolbar button), and the `<iframe>` embed snippet in HTML paste still works for those who intentionally want to embed a Google Doc.

## Related

Fixes #1082

Coverage: 72.34% (-0.01% vs `main`)

### DIFF
--- a/src/molecules/editor/extensions/iframe/iframe-allowlist.ts
+++ b/src/molecules/editor/extensions/iframe/iframe-allowlist.ts
@@ -31,8 +31,6 @@ export const IFRAME_ALLOWLIST: readonly string[] = [
   'codesandbox.io',
   'figma.com',
   'embed.figma.com',
-  'docs.google.com',
-  'drive.google.com',
   'notion.so',
 ]
 
--- a/src/molecules/editor/extensions/iframe/iframe-embed-utils.ts
+++ b/src/molecules/editor/extensions/iframe/iframe-embed-utils.ts
@@ -84,14 +84,6 @@ export const PLATFORM_CONFIGS: readonly PlatformConfig[] = [
     example: 'https://figma.com/design/…',
   },
   {
-    name: 'Google Docs',
-    ratio: 4 / 3,
-    defaultWidth: 600,
-    hosts: ['docs.google.com', 'drive.google.com'],
-    icon: 'lucide-file-text',
-    example: 'https://docs.google.com/document/d/…',
-  },
-  {
     name: 'Notion',
     ratio: 4 / 3,
     defaultWidth: 600,


netchampfaris:
@waterWang Better fix is to add a inline popup after pasting a embed link "Embed as iframe?", confirming that action should add the iframe embed, default behaviour should be paste as link.

PR #1091, opened 2026-08-14
fix: export defineModel props, add ContextMenuEmits, TabButtonsSlots, remove dead #target (Closes #1070)

### LINKED ISSUE #1070: Exported types are silent about props, emits and slots that exist
Part of #864. Four findings handed on by the cross-family vocabulary pass (#1054). Grouped because they are one defect wearing four hats: **the exported types do not describe the surface the components actually have.**

## Question

1. **`HoverCard`, `CommandPalette`, `KeyboardShortcutsModal` and `Tree` implement `open` / `expanded` through `defineModel` only.** The field never reaches their exported `*Props` interfaces, so the public type and the generated API table are both silent about a prop that exists and works.
2. **`ContextMenu` exports no `ContextMenuEmits` type**, though it fires `update:open`.
3. **`TabButtons` has no `defineSlots<>()`** — the only member of its family without one, so its slots are invisible to consumer type-checking and to `docs:gen`.
4. **`PickerShell` still declares a `#target` fallback slot.** Dead on the public surface (all three pickers forward only `#trigger`), and the last place in the repo spelling the retired name.

A consumer writing against the published types cannot see props that work, and the generated API tables — which the docs publish and CI checks for staleness (#917) — inherit the same blind spot.

## Not a release blocker

Every fix here is **additive**: adding a field to an exported interface, adding an emits type, adding `defineSlots`. Additive changes are allowed after the tag under P13, so this ships in a `1.0.x` minor rather than gating rc.1. Item 4 is a deletion, but of something already unreachable from the public surface.

Worth doing soon regardless: the API tables are what the docs show, and a table that omits a working prop reads as "unsupported".

waterWang:
Closes #1070

Four findings from the cross-family vocabulary pass, all confirmed by barista:

1. **`open` / `expanded` via `defineModel` only** — HoverCard, KeyboardShortcutsDialog, and Tree now declare their `open`/`expanded` prop in the exported `*Props` interface. Consumers can see and type-check these props.

2. **`ContextMenuEmits`** — added and exported the missing `ContextMenuEmits` type with the `update:open` event, matching the precedent set by `ContextMenuProps` which already declares `open`.

3. **`TabButtons` `defineSlots`** — added `TabButtonsSlots` type with `prefix` and `suffix` slot declarations, and wired `defineSlots<TabButtonsSlots>()` in the component. The generated API table will no longer show a truncated type blob.

4. **`#target` fallback slot removed from `PickerShell`** — the `#target` slot was the last place in the repo still spelling the retired name. None of the three pickers forward it, and the slot was unreachable from the public surface. Removed the template slot, the type declaration, and the two `slots.target` checks that treated it as a valid custom trigger.

### DIFF
--- a/src/components/ContextMenu/index.ts
+++ b/src/components/ContextMenu/index.ts
@@ -9,6 +9,7 @@ export type {
   ContextMenuItemSlots,
   ContextMenuOption,
   ContextMenuOptions,
+  ContextMenuEmits,
   ContextMenuProps,
   ContextMenuSlotFn,
   ContextMenuSlotProps,
--- a/src/components/ContextMenu/types.ts
+++ b/src/components/ContextMenu/types.ts
@@ -30,6 +30,10 @@ export interface ContextMenuProps {
   open?: boolean
 }
 
+export interface ContextMenuEmits {
+  'update:open': [open: boolean]
+}
+
 export type ContextMenuSlots = Omit<MenuSlots, 'default' | 'trigger'> & {
   /** The right-clickable region that opens the menu. */
   default?: (props: ContextMenuTriggerSlotProps) => any
--- a/src/components/HoverCard/types.ts
+++ b/src/components/HoverCard/types.ts
@@ -44,6 +44,9 @@ export interface HoverCardProps {
    * surface.
    */
   arrow?: boolean
+
+  /** Controls the visibility of the hover card. */
+  open?: boolean
 }
 
 /** Slot props passed to the `#trigger` slot. */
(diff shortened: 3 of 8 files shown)


netchampfaris:
Merged into #1133 and finished there, so this can close when that lands.

Two things needed fixing once the API tables were regenerated against it. `PickerShell.vue` was left with a stray `</slot>` after the `#target` opening tag was deleted — Vue's parser drops it with a warning, so `vue-tsc` never complained. And `TabButtonsSlots` was added but never exported, unlike every other `*Slots` type. The dead `#target` slot needed no changelog entry: `changelog.md:1335` already documents it as removed, so the code was contradicting the published docs.

PR #1093, opened 2026-08-14
fix(slots): re-evaluate slot-derived computeds when a slot toggles

netchampfaris:
Reported in the Raven `Raven-frappe-ui` channel: the prefix story on the TextInput docs page is broken. That turned out to be one instance of a defect that runs through the library, so this fixes all of them.

## What was broken

On https://ui.frappe.io/docs/components/textinput, turning on the playground's `prefix` switch paints the icon on top of the input text. The same happens for `suffix`. Any consumer that puts a `#prefix` behind a `v-if` hits it.

The static "Prefix and suffix slots" preview looks right, because those slots are filled at mount.

## Root cause

`TextInput` builds its padding classes in a `computed` that reads `useSlots()`:

```ts
slots.prefix ? 'ps-8' : 'ps-2'
```

`useSlots()` returns `instance.slots`. Vue mutates that object in place and does not track it, so the computed caches whichever slots were filled at mount and never re-runs. The prefix element itself renders (the template reads `$slots` at render time), but the input keeps its slot-less `ps-2` and the two overlap.

`hasLabeling` in the same file had the same defect for the `#label` and `#description` slots.

`useInputLabeling` already fixed this for `labelledBy` / `describedBy` in #1079, with a ref bumped in `beforeUpdate`. `beforeUpdate` runs after `updateSlots` and before the render function, so it is the invalidation the computeds are missing.

## Not a docs-only problem

The playground toggles the slot the ordinary way, `<template v-if="values.prefix" #prefix>`. Consumers do the same. Gameplan's `Page.vue` toggles `#suffix` on `PageHeaderMobile` when the page document arrives from the server, and the more-options menu never appears.

Two other Vue libraries shipped the same defect in the same shape: Element Plus (element-plus#18978, fixed by moving the check into the template) and Nuxt UI (still present in `Input.vue` at v4.11.0).

## Why this shape

Vue's own docs prescribe the trigger:

> the properties of `attrs` and `slots` are **not** reactive. If you intend to apply side effects based on changes to `attrs` or `slots`, you should do so inside an `onBeforeUpdate` lifecycle hook.

Nuxt UI ships exactly that in `Page.vue` (`shallowRef` seeded from the slots, refreshed in `onBeforeUpdate`), after nuxt/ui#4947 and nuxt/ui#4969. They applied it to one component and left three others broken in two incompatible shapes. Factoring it once avoids that.

Vue core wraps `instance.slots` in the same kind of proxy in `getSlotsProxy`, but only in dev builds, and only HMR triggers the matching dep. The behaviour is still open upstream as vuejs/core#11227.

Libraries that author in render functions (Vuetify, Naive UI, Ant Design Vue, reka-ui, Headless UI) never hit this, because a render function re-reads the slots on every update. frappe-ui is `<script setup>` SFCs, so that escape is not available.

## What changed

- New `useSlotTick()`: the invalidation, lifted out of `useInputLabeling` so it
  has one home. It memoizes one ref per component instance, so a component
  reading it from two places registers one `beforeUpdate` hook.
- New `useReactiveSlots()`: `useSlots()` wrapped in a proxy that reads the tick
  on every access. A `computed` that reads the returned object subscribes by
  itself. Both stay internal, neither is exported from `src/index.ts`.
- 28 components take their slots from it, with the declared slot names kept on
  the handle so a typo still fails to compile.
- `PropsTable.vue` in `vitepress/` reads its `#code` slot in the template
  instead, which needs no composable.
- `CONTEXT.md` states the rule.

Three distinct shapes needed fixing, not one:

1. a `computed` or `watch` getter in the same file (most components),
2. the slots object handed to a child as a prop (`Dropdown` to `MenuItemContent`),
3. the slots object handed to descendants through `provide` (`ListView`).

No prop, slot or event changed.

## Verification

`src/components/shared/slotReactivity.cy.ts` covers 13 components. Every case mounts with the slot empty, fills it, and empties it again.

With the tick reads stripped out of the proxy, all 13 fail. That red proof caught three cases that passed for the wrong reason and were rewritten.

```
TextInput.cy.ts        19 passing
slotReactivity.cy.ts   13 passing
```

Local run on the rebased head: full Cypress component suite, `yarn test` 1585 passing across 95 files, `yarn type-check` clean, `yarn docs:check` clean.

## Known limitation

`<template #prefix><span v-if="show" /></template>` still reserves the prefix region when `show` is false. The slot key never disappears, so no presence check catches it. Naive UI and Ant Design Vue solve that by invoking the slot and filtering `Comment` vnodes. Different defect, present before this PR, not addressed here.

Docs preview: https://ui.frappe.io/pr-preview/pr-1093/

Coverage: 72.52% (+0.09% vs `main`)

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -25,7 +25,9 @@ live in `src/components/` or a domain dir (e.g. `frappe/` for Frappe-integrated 
 A component that reads its own slots outside the template takes them from
 `useReactiveSlots()`, never `useSlots()`. Vue mutates `instance.slots` in place
 and does not track it, so a `computed` over `useSlots()` caches whichever slots
-were filled at mount. Reading `$slots` in a template is already correct.
+were filled at mount. Reading `$slots` in your own template is already correct;
+handing the object to a child (as a prop, or through `provide`) is not, because
+the child re-reads it only when something else re-renders the child.
 
 ## Lifecycle & control
 
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -21,6 +21,12 @@ A component that composes atoms into a higher-level control — e.g. `Link` (com
 `Combobox`). Follows the same design rules as atoms (P5 labeling, P10 styling, …). May
 live in `src/components/` or a domain dir (e.g. `frappe/` for Frappe-integrated controls).
 
+**slots (reading your own)**:
+A component that reads its own slots outside the template takes them from
+`useReactiveSlots()`, never `useSlots()`. Vue mutates `instance.slots` in place
+and does not track it, so a `computed` over `useSlots()` caches whichever slots
+were filled at mount. Reading `$slots` in a template is already correct.
+
 ## Lifecycle & control
 
 **open**:
--- a/experimental/CodeEditor/CodeEditor.vue
+++ b/experimental/CodeEditor/CodeEditor.vue
@@ -46,13 +46,12 @@ import {
   onMounted,
   ref,
   useAttrs,
-  useSlots,
   watch,
 } from 'vue'
 import type { EditorView } from '@codemirror/view'
 import type { Compartment, Extension } from '@codemirror/state'
 import { useInputLabeling } from '../../src/composables/useInputLabeling'
-import { useSlotTick } from '../../src/composables/useSlotTick'
+import { useReactiveSlots } from '../../src/composables/useReactiveSlots'
 import InputLabel from '../../src/components/InputLabeling/InputLabel.vue'
 import InputDescription from '../../src/components/InputLabeling/InputDescription.vue'
 import InputError from '../../src/components/InputLabeling/InputError.vue'
@@ -79,8 +78,7 @@ defineSlots<{
 }>()
 
 const attrs = useAttrs()
-const slots = useSlots()
-const slotTick = useSlotTick()
+const slots = useReactiveSlots()
 
 const {
   labelId,
@@ -103,12 +101,11 @@ const {
 
 // Render the labeling chrome (and its wrapping div) only when something needs
 // it — otherwise the editor mounts bare, preserving the primitive's footprint.
-const hasLabeling = computed(() => {
-  slotTick.value
-  return Boolean(
+const hasLabeling = computed(() =>
+  Boolean(
     props.label || slots.label || rendersDescription.value || hasError.value,
-  )
-})
+  ),
+)
 
 const el = ref<HTMLElement | null>(null)
 
--- a/experimental/CodeEditor/CodeEditor.vue
+++ b/experimental/CodeEditor/CodeEditor.vue
@@ -52,6 +52,7 @@ import {
 import type { EditorView } from '@codemirror/view'
 import type { Compartment, Extension } from '@codemirror/state'
 import { useInputLabeling } from '../../src/composables/useInputLabeling'
+import { useSlotTick } from '../../src/composables/useSlotTick'
 import InputLabel from '../../src/components/InputLabeling/InputLabel.vue'
 import InputDescription from '../../src/components/InputLabeling/InputDescription.vue'
 import InputError from '../../src/components/InputLabeling/InputError.vue'
@@ -79,6 +80,7 @@ defineSlots<{
 
 const attrs = useAttrs()
 const slots = useSlots()
+const slotTick = useSlotTick()
 
 const {
   labelId,
@@ -101,11 +103,12 @@ const {
 
 // Render the labeling chrome (and its wrapping div) only when something needs
 // it — otherwise the editor mounts bare, preserving the primitive's footprint.
-const hasLabeling = computed(() =>
-  Boolean(
+const hasLabeling = computed(() => {
+  slotTick.value
+  return Boolean(
     props.label || slots.label || rendersDescription.value || hasError.value,
-  ),
-)
+  )
+})
 
 const el = ref<HTMLElement | null>(null)
 
--- a/experimental/CodeEditor/CodeEditor.vue
+++ b/experimental/CodeEditor/CodeEditor.vue
@@ -40,14 +40,7 @@
 // tree-shakes into its own async chunk — apps that never render a code field pay
 // no runtime cost (importing `frappe-ui/experimental` pulls in no editor code
 // until a field actually mounts).
-import {
-  computed,
-  onBeforeUnmount,
-  onMounted,
-  ref,
-  useAttrs,
-  watch,
-} from 'vue'
+import { computed, onBeforeUnmount, onMounted, ref, useAttrs, watch } from 'vue'
 import type { EditorView } from '@codemirror/view'
 import type { Compartment, Extension } from '@codemirror/state'
 import { useInputLabeling } from '../../src/composables/useInputLabeling'
--- a/experimental/CodeEditor/CodeEditor.vue
+++ b/experimental/CodeEditor/CodeEditor.vue
@@ -63,7 +63,7 @@ const props = withDefaults(defineProps<CodeEditorProps>(), {
   size: 'md',
 })
 const emit = defineEmits<CodeEditorEmits>()
-defineSlots<{
+const declaredSlots = defineSlots<{
   /** Overrides the rendered label content. Receives `{ required }`. */
   label?: (props: { required: boolean }) => any
   /** Overrides the rendered description content. */
@@ -71,7 +71,7 @@ defineSlots<{
 }>()
 
 const attrs = useAttrs()
-const slots = useReactiveSlots()
+const slots = useReactiveSlots<typeof declaredSlots>()
 
 const {
   labelId,
--- a/experimental/ListView/ListView.vue
+++ b/experimental/ListView/ListView.vue
@@ -24,7 +24,8 @@ import ListHeader from './ListHeader.vue'
 import ListRows from './ListRows.vue'
 import ListGroups from './ListGroups.vue'
 import ListSelectBanner from './ListSelectBanner.vue'
-import { ref, reactive, computed, provide, watch, useSlots } from 'vue'
+import { ref, reactive, computed, provide, watch } from 'vue'
+import { useReactiveSlots } from '../../src/composables/useReactiveSlots'
 
 defineOptions({
   inheritAttrs: false,
@@ -60,7 +61,7 @@ const props = defineProps({
   },
 })
 
-const slots = useSlots()
+const slots = useReactiveSlots()
 
 let selections = reactive(new Set())
 let activeRow = ref(null)
--- a/experimental/MultiEmailInput/MultiEmailInput.vue
+++ b/experimental/MultiEmailInput/MultiEmailInput.vue
@@ -31,6 +31,7 @@ import {
   LabelingWrapper,
 } from '../../src/components/InputLabeling'
 import { useInputLabeling } from '../../src/composables/useInputLabeling'
+import { useSlotTick } from '../../src/composables/useSlotTick'
 import { usePortalTarget } from '../../src/composables/usePortalTarget'
 import {
   inputFontSizeClasses,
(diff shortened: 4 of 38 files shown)


netchampfaris:
Confirmed the sibling finding, declining it here as scope.

Checked each one: `Textarea.vue:104`, `Select.vue:111`, `Rating.vue:385`, `Slider.vue:209`, `RadioGroup.vue:113`, `Radio.vue:73` and `Switch.vue:167`/`:216` all read `slots.label` / `slots.description` inside a plain `computed`. Same defect, and `useSlotTick()` is now one line each.

Keeping it out of this PR: it originates from a bug report about the TextInput prefix, and none of those components was reported broken. Seven more components plus a toggle test each triples the diff for a defect nobody has hit yet. Better as its own PR, where the tests get room.

netchampfaris:
Reversing the scope call above — the sweep is now in this PR.

A parser over every `computed` body in `src` and `experimental` found 22 sites in 21 components, more than the follow-up list. Beyond the labeling and prefix/suffix sets already named: `Button` (caches the vnodes it inspects for an icon-only default slot), `NumberCard`, `DatePicker`, `DateTimePicker`, `PickerShell`, `ThemeSwitcher`, `Pill`, `TabTrigger`, `Combobox` (`isButtonMode`).

New spec `src/components/shared/slotReactivity.cy.ts` covers eight, one per distinct symptom. All eight fail with the invalidation removed from `useSlotTick` and pass with it restored.

/barista review

netchampfaris:
Round 4. Took barista's `useReactiveSlots()` suggestion, plus the two smaller nits.

- `useReactiveSlots()` is a proxy over `useSlots()` that reads the tick on every access. Reading the slots *is* the subscription, so all 24 bare `slotTick.value` statements are gone and component 25 cannot forget it.
- `useSlotTick()` memoizes one ref per component instance, keyed on the instance in a `WeakMap`. The ~10 labeling components no longer register two `onBeforeUpdate` hooks.
- Both stay internal. Neither is exported from `src/index.ts`.
- `Button`'s `defaultSlot` gate survives the rewrite: the computed's value is the slot function itself, so it only propagates when the slot changes, not on every render.
- Formatted the five files this branch made non-conforming. The seven that already failed `prettier --check` on `main` are untouched.

Red proof re-run against the proxy: strip the tick reads out of `useReactiveSlots` and all 10 cases in `slotReactivity.cy.ts` fail, plus the TextInput toggle test. Restored, the full component suite is 1301/1301 across all 82 specs with no flakes this run. `yarn test` 1529 passing, `yarn type-check` clean on every touched file.

netchampfaris:
Round 5. Both nits fixed.

**Typed slot names.** Right catch — I regressed those eight. `useReactiveSlots<AlertSlots>()` where a named type exists (`Alert`, `SidebarCard`, `NumberCard`, `DatePicker`, `DateTimePicker`), and `useReactiveSlots<typeof declaredSlots>()` where `defineSlots` takes an inline literal (`TabTrigger`, `PickerShell`, `Pill`). Proof the guard is back — temporarily misspelling the key in `Alert`:

```
src/components/Alert/Alert.vue(75,17): error TS2551: Property 'actionz'
does not exist on type 'AlertSlots'. Did you mean 'actions'?
```

Left alone: the 15 that came from `useSlots()`. Several declare `defineSlots<{...}>()` without a binding and could be typed the same way, but they were untyped before this PR too, so that is a separate improvement rather than a regression to undo here.

**Discoverability.** Added to `CONTEXT.md`, under Composition:

> **slots (reading your own)**: A component that reads its own slots outside the template takes them from `useReactiveSlots()`, never `useSlots()`. Vue mutates `instance.slots` in place and does not track it, so a `computed` over `useSlots()` caches whichever slots were filled at mount. Reading `$slots` in a template is already correct.

No ESLint rule — the repo has no ESLint config to hang one on, and adding that setup is well outside this PR.

`slotReactivity` 10/10 plus `Alert`, `NumberCard`, `DatePicker`, `DateTimePicker`, `Tabs`: 134/134. `yarn type-check` clean on every touched file.

netchampfaris:
Round 6. Both are real and both are fixed — good catches, and my sweep was at fault twice over.

- `Dialog.vue:200` — `allSlots` is now `useReactiveSlots()`.
- `Dropdown.vue:76` and `ContextMenu.vue:52` — same. That closes the `MenuItemContent` chain for the two sources that were still plain.

Why the parser missed them:

1. The pattern required a handle named exactly `slots`, so `allSlots.title` never matched. Dialog is the only file with a second handle.
2. It looked for the read in the same file as the binding. `MenuItemContent` reads `props.slotFns`, in a child, under a different name — nothing in Dropdown or ContextMenu to match on.

Re-ran a widened sweep: bind every handle assigned from `useSlots` / `useReactiveSlots` / `defineSlots` whatever its name, plus slots objects arriving as a prop, then look for reads inside `computed` bodies. It returns exactly your two findings and nothing else. `MultiSelectResults` and `ComboboxResults` also touch `props.slotFns`, but from plain functions called by the template, which re-run on render.

Two more cases in `slotReactivity.cy.ts`, at 12 now. Both were non-trivial to make honest:

- The Dropdown menu must stay open across the toggle. Closing it remounts `MenuItemContent` and rebuilds its computeds, so the test passes with the bug present. A click anywhere outside dismisses the menu, so the slot is toggled from a ref rather than a button.
- The Dialog toggle needs `force: true` — an open modal sets `pointer-events: none` on `body`.

Red proof with the tick reads stripped out of `useReactiveSlots`: 0 passing, 12 failing. Restored: 12/12, and `Dialog` 24/24, `Dropdown` 17/17, `ContextMenu` 14/14 alongside. `yarn type-check` clean on all three.

netchampfaris:
Round 7. `ListView.vue:64` is now `useReactiveSlots()`.

You are right that this is a third shape, and it is the one that breaks the CONTEXT.md rule I wrote last round. Tightened it:

> Reading `$slots` in your own template is already correct; handing the object to a child (as a prop, or through `provide`) is not, because the child re-reads it only when something else re-renders the child.

The three shapes are now in the PR body: same-file `computed`, slots-as-a-prop (Menu), slots-through-`provide` (ListView). The proxy covers all three for the same reason — the subscription attaches wherever the read happens, not where the object was made.

New case in `slotReactivity.cy.ts`, 13 now. It first passed with the bug present: my harness built `rows` and `columns` as fresh array literals per render, which changes the provided computed's deps and re-renders the descendants regardless. Hoisted both out so the slot toggle is the only thing that moves. Red proof after that: 0 passing, 13 failing.

Agreed on `Checkbox.vue:96` — getters into `useInputLabeling`, which reads the tick itself. No change.

Full component suite on the previous head (`cb23588b2`): 1303/1303 across 82 specs, no flakes.

netchampfaris:
Round 8. Both done.

- `Tabs.vue:44` — `useReactiveSlots<typeof declaredSlots>()`. Widened the sweep to `watch`, `watchEffect`, `watchPostEffect` and `watchSyncEffect` as well as `computed`; Tabs is the only site, and `MenuItemContent` is the only other hit, already covered through its four sources.
- `useReactiveSlots.ts` doc comment now carries the prop/`provide` caveat, matching CONTEXT.md.

No test for the Tabs warning, and this one I want to flag rather than bury. The watch drives nothing but a `console.warn`, so the only observable is the console. I wrote the spec, and with the fix neutered it does not fail in four seconds — it hangs the runner outright, because `cy.spy(console, 'warn')` waiting on a warning that never comes takes Cypress with it. A case I cannot red-prove is not worth shipping, so the fix goes in without one.

`slotReactivity` 13/13 and `Tabs` 42/42. Local full suite on the previous head (`7b238bb18`): 1304/1304 across 82 specs, and CI was green on it too.

netchampfaris:
Round 9. Both done.

- `DateRangePicker.vue:145` — binding dropped, the macro stays. It was the last `const slots = defineSlots(...)` in the repo.
- Added a **Cost** section to the PR body naming the trade you asked about: the tick bumps on every update, so a computed over the proxy re-runs per update of its component. Cheap for the boolean checks; `MenuItemContent`'s four content computeds do re-materialize vnodes on each Dropdown update, and `Button` is gated because there the cost fell on every render of the most-mounted component in the library.

`DateRangePicker` 22/22, `yarn type-check` clean.

netchampfaris:
Round 10. Both done.

- `Dialog.vue` — went further than dropping the binding. `allSlots` takes the name `slots`, and the macro loses its binding, so there is one handle instead of two four lines apart. Typed as `DialogSlots`, so the coin flip you describe is now a type error either way.
- **Cost** section extended with the `provide` fan-out: every row's render effect subscribes to `ListView`'s tick, so a `ListView` update re-renders every row, and that is the one place where the cost scales with data rather than component count.

`Dialog` 24/24, `slotReactivity` 13/13, `yarn type-check` clean.

netchampfaris:
Round 11. Done, and wider than the ask.

You named four. There are sixteen: six with a named `*Slots` type already imported and unused (`Combobox`, `MultiSelect`, `Dropdown`, `ContextMenu`, `Select`, `MultiEmailInput`), and ten declaring an inline literal, which take the `declaredSlots` + `typeof` pattern. You said the inline ones were worth doing eventually and would not hold the PR — but it is the same one-line edit, so booking a follow-up for it made no sense.

Two needed more than a type argument. `Combobox` and `MultiSelect` forward their slots to a Results child that dispatches on dynamic names (`item-${slot}`), which an enumerated type cannot express — `Type 'ComboboxSlots' is not assignable to type 'SlotFns'`. Those cast once at the forward, with a comment, and keep the typed handle for their own reads. Same object, so reads still go through the proxy.

`ListView` stays untyped: it has no `defineSlots` to take a type from.

Guard proven at both shapes:

```
src/components/Combobox/Combobox.vue(205,23): error TS2551: Property 'triggerr'
does not exist on type 'ComboboxSlots'. Did you mean 'trigger'?

src/components/TextInput/TextInput.vue(181,13): error TS2551: Property 'prefixx'
does not exist on type 'Readonly<__VLS_Slots> & __VLS_Slots'. Did you mean 'prefix'?
```

Full component suite: 1304/1304 across 82 specs. `yarn test` 1529 passing. `yarn type-check` produces no error not already on `main`.

netchampfaris:
Round 12. Converted, for the reason you give: a carve-out that lives only in a PR description is not a rule.

`Checkbox.vue:97` is `useReactiveSlots<typeof declaredSlots>()`. It worked before — the getters run inside `useInputLabeling`'s tick-subscribed computeds — but `Boolean(slots.label)` is a read outside the template, and CONTEXT.md now says those take the proxy. `useSlots()` appears in exactly one place, the composable that wraps it.

`Checkbox` 21/21, `slotReactivity` 13/13, `yarn type-check` produces no error not already on `main`, prettier clean.

And CI finished green on the previous head (`9b43af698`), so the sixteen type arguments are no longer local-only: `Type Check`, `test`, all three Cypress shards, `API Tables`, `build-preview` and `coverage-report` all pass.

netchampfaris:
Round 15. Rebased onto main and closed one more site.

The branch had gone stale: main moved 68 commits and `Button.vue` conflicted with `5c0d65ad2 refactor(Button): drop the exposed rootRef`. Resolved by keeping main's `setup(props, { attrs, slots: rawSlots })` signature without `expose`.

Two new commits:

- `fix(ThemeSwitcher): point the composable import at the package alias` — main moved ThemeSwitcher from `src/components/` to `experimental/`, which broke the relative path this branch wrote. `yarn type-check` caught it. The file already imports `useColorScheme` and `warnDeprecated` through the `#` subpath imports, so `useReactiveSlots` now does the same.
- `fix(docs): PropsTable reads its #code slot in the template` — the last site, missed because it lives in `vitepress/` rather than `src/`. `hasCustomCodeSlot` was a computed over `useSlots()`. No caller toggles `#code`, so nothing was visibly broken, but it contradicted the rule this PR writes into `CONTEXT.md`. Fixed with a template read, which needs no composable and no new export.

A fresh sweep of the rebased tree finds `useSlots()` in exactly one place, `useReactiveSlots.ts`. The two hits in `ComboboxResults.vue` and `MultiSelectResults.vue` are doc comments.

Local on `7a4c97882`: Cypress 1317/1317 across 84 specs, `slotReactivity.cy.ts` 13/13, `TextInput.cy.ts` 19/19. `yarn test` 1585 passing. `yarn type-check` clean, which it was not before this rebase. `yarn docs:check` clean.

The PR description is rewritten to cover the whole change rather than just the TextInput symptom, and now records a known limitation: a slot whose key stays but whose body is `v-if`'d out still reserves its region. That is a different defect, present before this PR.

PR #1095, opened 2026-08-14
ci: type-check the library on every PR, and fix the config that hid 38 errors

### LINKED ISSUE #1074: CI never type-checks the library, so the type half of the migration story is unverified
Part of #864. Found by barista while reviewing #1072.

## Question

**Nothing in CI type-checks this library.** `package.json:31` defines `type-check` (`vue-tsc --noEmit -p tsconfig.app.json`), but `.github/workflows/tests.yml` runs vitest, the API-table staleness check (#917) and Cypress — never `vue-tsc`. No hook runs it either.

That matters more at `1.0.0` than it did in beta, because a large share of the migration story is *carried by types*:

- The Badge removal (#1069) is explicitly "loud for TypeScript, silent for JS" — the compile error **is** the signal an app gets. Nothing in this repo verifies that error still fires.
- The vocabulary renames (#1054) leave the guide as the whole migration path for Tabs, and type errors as the loud half for the rest.
- ADR-0012 freezes the v2 composable surface; a type regression widening or narrowing it would not be caught.
- Exported types were already found lying about the surface (#1070).

A union can regress, an exported type can drift, and every automated check here stays green.

## The wrinkle that makes this more than one line of YAML

`main` currently reports a **~77-error baseline**, all in `vitepress/`, `src/mocks/handlers.ts`, `src/components/Menu/`, the editor extensions and `experimental/`. So `yarn type-check` cannot simply be added as a required step — it fails on a clean checkout today.

Decide which:

1. **Fix the 77**, then gate on zero. Cleanest end state, unknown size — the count has been stable across every branch measured today, which suggests they are structural rather than accumulating.
2. **Gate on "no new errors vs the merge base"**, the way every agent working this map has been checking by hand: run the baseline, run the branch, diff the per-file error lists. Catches regressions immediately without blocking on the cleanup, and the cleanup can then proceed incrementally.
3. **Narrow the project** so the checked scope excludes the known-bad directories, and gate on zero for the rest.

Option 2 mirrors what #917 did for API tables: fail on *drift*, not on an absolute standard the repo does not meet yet.

Not a release blocker: this catches future regressions rather than fixing a defect in the shipped surface, and the tag's own type correctness has been verified by hand on every PR in this map.

netchampfaris:
Closes #1074.

## What was wrong

Two things, and the second is why the first was never noticed.

**Nothing in CI ran `vue-tsc`.** `tests.yml` ran vitest, the API-table check and Cypress. A type regression could widen or narrow the frozen `1.0.0` surface with every check green.

**`tsconfig.app.json` set `baseUrl: "."`.** Every bare specifier resolved against the repo root first, so `import { useData } from 'vitepress'` and `import type { Plugin } from 'vite'` hit our own `vitepress/` and `vite/` directories instead of the npm packages. That is 38 of the 78-error baseline the ticket describes — phantom errors against modules the code never meant to import.

The `exclude` list was wrong in both directions. `src/mocks/**/*` and `src/**/stories/**/*` were listed, but exclusion does not remove a file an included file imports — the tests import the mocks, so the mocks were checked anyway. Stories were genuinely unchecked, which is how `FormControl.story.vue` kept importing `src/components/Story.vue` for two years after `8bb3c5a14` deleted it.

`tsconfig.node.json` extended nothing, so it ran at the ES5 default with lib checking on: 8 errors, 7 of them inside `node_modules`.

## The decision

Option 1 from the ticket — **fix to zero and gate on zero** — not option 2's drift diff. Once the config stops shadowing packages the baseline is zero, so there is nothing left for a drift diff to tolerate.

**Stories join the program.** They are the docs' live examples and the source `docs:gen` reads. They cost 27 errors, all real, all fixed.

**Cypress specs stay out.** `cy.mount()`'s component generics produce ~202 errors that are a cleanup of their own, unrelated to the shipped type surface.

## Numbers

| | errors |
|---|---|
| `main`, as measured | 78 |
| after dropping `baseUrl` | 40 |
| after adding stories | 67 |
| after this PR | **0** |

## Fixes behind the count

- **Menu render helpers** take `VNodeChild`, which is what `MenuSlotFn` is declared to return. Internal; `src/utils/vnode.ts` is unreachable from any entry point.
- **`experimental/Charts`** — `ECharts` imported as a type instead of a bare `echarts.` namespace; `series: []` annotated so it stops inferring `never[]`.
- **`src/mocks/handlers.ts`** — msw path params are `string | readonly string[] | undefined`; two helpers were untyped.
- **`vitepress/`** — the sidebar comes from `defineDocsConfig` in a shape VitePress' own `Sidebar` union does not describe; `@types/markdown-it` ships `state_core.d.mts`, which bundler resolution only reaches through the `.mjs` specifier.
- **Eight stories** passing values their component rejects, including `FileUploader`'s `@success` handler typed as `File` when the payload is `UploadedFile`.
- **`FormControl.story.vue` deleted** — the only `*.story.vue` left, importing a component deleted in `8bb3c5a14`, referenced by nothing.

## One defect found, not fixed here

`update:open` and `update:query` are declared **twice** on five components — once via `defineModel`, once in the exported `*Emits` interface. The merge loses the type, so `@update:open="x = $event"` hands a consumer `unknown`. Confirmed by probe: removing the duplicate key restores the type.

`Combobox`, `MultiSelect`, `Dropdown`, `Select`, `SettingsDialog` — 7 emits. Three call sites in this repo carry a cast with a comment rather than a fix, because removing a member from an exported interface needs its own decision and overlaps #1070 / #1091. Filed separately.

## Verification

- `yarn type-check` — exit 0, both projects
- `yarn test` — 1529/1529
- `yarn build` — clean
- `yarn docs:check` — tables match
- Cypress `Dropdown` + `ContextMenu` + `Combobox` — 104/104 (the specs that exercise the widened Menu renderers)

Docs preview: https://ui.frappe.io/pr-preview/pr-1095/

Coverage: 72.46% (+0.09% vs `main`)


netchampfaris:
All four taken, [3f93e7d](https://github.com/frappe/frappe-ui/commit/3f93e7d746d217598e3b959dcbb18360749118cb).

**`vnode.ts` behaviour, now tested** — you were right that it is not type-only, and the two changes are narrower than the write-up suggested. A bare **number** was already `true` before: the old code reached `5.type`, found `undefined`, and fell through to `return true`. What actually moved is a blank bare **string** (`'   '` was `true`, now `false` — the old code never reached a `.trim()`) and a **Fragment with string children** (`false` → `true`, because the old `Array.isArray(node.children) ? … : []` threw the string away). Both are the right answer, neither had a test.

`src/utils/vnode.test.ts` — 12 tests over `null` / `false` / blank string / number / nested array / Fragment-with-string-children / comment / blank text vnode, for both helpers. **Mutation-checked**: restoring either old branch turns exactly three of the twelve red, so they are not vacuous.

**`series`** — `EChartsOption['series']`, not `any[]`. Agreed on the reasoning: `useAxisChartOptions` is reachable from `frappe-ui/experimental`, so `any[]` would have stopped checking consumers' series lists.

**Trailing newline** restored, and noted why it would have stayed missing: `yarn format` is scoped to `./src`.

**Casts now name #1096** — [Model emits are declared twice, so `$event` is unknown for consumers](https://github.com/frappe/frappe-ui/issues/1096), filed with the probe evidence and the five affected components. All three sites point at it, so they come out with the fix.

Verified after the changes: `yarn type-check` exit 0 · `yarn test` 1541/1541 (93 files).

/barista review

netchampfaris:
Taken, [3ac451c](https://github.com/frappe/frappe-ui/commit/3ac451ce6dd923623f1570a33bb1916c9216723d) — but the else branch is **not** dead, which changes which of your two options is right.

`Sidebar.vue:15` fills that state from `theme.sections ?? theme.sidebar ?? []`. `sections` is ours and is `SidebarSection[]`; **`sidebar` is VitePress' own `DefaultTheme.Sidebar`, which admits a flat list of links with no `items`.** `getSidebarList` returns its input unchanged while declaring `SidebarSection[]`, so the narrow type is the lie, not the branch.

So: your second option, scoped to the one file rather than widening `state.sidebarList` and forcing narrowing into `Sidebar.vue` and `MobileNavSheet.vue` too.

```ts
const entries = state.sidebarList as Array<SidebarSection | SidebarItem>

const linkInfos = entries.reduce<SidebarItem[]>((acc, cur) => {
  'items' in cur ? acc.push(...cur.items) : acc.push(cur)
  return acc
}, [])
```

One cast, sitting on the declaration that is actually wrong, with a comment saying so; `'items' in cur` does real narrowing, so neither branch is a lie now. Runtime behaviour unchanged.

Worth recording rather than fixing here: **that flat shape already breaks the theme two files over.** `MobileNavSheet.vue:20` calls `section.items.filter(...)` unguarded, so a flat `theme.sidebar` throws there, and `Sidebar.vue:44` renders nothing. Making the docs theme genuinely support a flat sidebar is a change to the docs data model, not to a type-check gate — out of scope for this PR, and it is `frappe-ui/vitepress`, which ships with no stability promise under P14.

`yarn type-check` exit 0 after the change.

/barista review

netchampfaris:
Comment sharpened, [07cc7c0](https://github.com/frappe/frappe-ui/commit/07cc7c0c17d10b1f5e9bbf5b8930bdcfa84b01d6) — you are right, and "the declaration is a lie" was the wrong framing.

Checked both lines: `sidebar: SidebarSection[]` is required at `index.node.ts:51`, and `:167-168` writes `sidebar` and `sections: sidebar` from that one value. So **through `defineDocsConfig` the declared type is correct**, and the flat shape is reachable only by a consumer hand-writing raw `themeConfig`, which bypasses the helper. The comment now says exactly that, so it no longer contradicts `CommandPalette.vue:27`.

That also settles the `MobileNavSheet.vue:20` / `CommandPalette.vue:29` unguarded `.items` reads: they are safe on every path this repo's own helper produces, and only the bypass path could reach them. Leaving both as they are — supporting a hand-written flat sidebar is a docs data-model decision, not a type-check gate, and `frappe-ui/vitepress` ships with no stability promise under P14.

No further round needed on my side either. Doc-comment only; `yarn type-check` exit 0.

PR #1097, opened 2026-08-15
refactor(list): finalize the v1 CSS styling-hook contract (ADR-0017)

netchampfaris:
## What this does

Freezes the `frappe-ui/list` CSS-variable surface before v1: the public hooks are exactly `--list-columns`, `--list-gap` and `--list-row-padding-x`, plus the preset sugar `list-cols-[…]` / `list-gap-*` / `list-row-px-*`. Everything else is explicitly internal.

- **Uniform ancestor theming.** The `:where([data-slot='list'])` root declaration is gone; rows and headers read `var(--list-columns, var(--_list-columns-default, auto minmax(0,1fr) auto))` at the use site. All three hooks now behave identically — settable on the `List` or inherited from any ancestor to theme a subtree — where previously `--list-columns` silently shadowed ancestor values. A consumer class still beats the `columns` prop through the internal carrier.
- **Internal vars take the `--_list` prefix**: `--_list-columns-default`, `--_list-checkbox-width`, `--_list-row-height`. None was documented API. Row height stays prop-only on purpose — a per-breakpoint height var would silently desync `virtual` windowing, so the `rowHeight` JSDoc no longer advertises a var name.
- **Docs**: the Styling hooks section now covers all three hooks, the deliberately asymmetric `--list-row-padding-x` default (`0.75rem` on interactive rows, flush header) and the `list-row-px-3` pairing that aligns them, the props-vs-hooks rule (behavior → props, pure geometry → vars), the `--_list` = internal rule, and three previously undocumented slots (`list-group`, `list-group-header`, `list-header-checkbox`) with "slots not listed are internal."
- **Story**: `Columns.vue` becomes the canonical interactive table — `v-model:active` + `list-row-px-3`, so the shipped demo actually exercises the header/row alignment contract. The per-row Remove `Button` had to go: activatable rows render as `<button>`, and a nested button is invalid HTML.
- **ADR-0017** records the seven conventions (unprefixed hooks, `--_` internals, use-site fallbacks, `:where()`, paint-vs-behavior, preset sugar, no `@property`) and the rejected alternatives, so the next family with CSS knobs copies this shape.
- **Tests**: a new `List (styling hooks)` block in `List.cy.ts` pins hook-beats-prop, ancestor inheritance for all three hooks, the dual inset default, preflight not eating button-row padding, and the sugar utilities.
- **Changelog**: entry under Unreleased for the two observable changes (ancestor theming; internal renames).

## Verification

- vitest: 92 files / 1527 tests pass; the single runner error is a pre-existing editor `MediaNodeView` issue, identical on the untouched tree.
- `vue-tsc`: same 78 pre-existing errors before and after, none in list files.
- `yarn docs:gen`: regenerated cleanly; only `list.api.md` changed (tracking the JSDoc edits).
- The new Cypress specs did not run locally (the sandbox's network policy blocks the Cypress binary download) — CI is the authoritative run for them.

https://claude.ai/code/session_01Hz2gWNm6DktAH61BAvMuBM

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hz2gWNm6DktAH61BAvMuBM)_

Docs preview: https://ui.frappe.io/pr-preview/pr-1097/

Coverage: 72.50% (±0.00% vs `main`)


netchampfaris:
Both review findings were real and are addressed in a42ab15:

**Carrier leak into nested lists** (Greptile P1) — valid, though pre-existing rather than introduced here: the identical inheritance chain ran through the old unprefixed names (`--list-columns-default`, `--list-checkbox-width`, `--list-row-height`) before the rename. Fixed properly while freezing: every list root now resets the prop carriers to `initial` (guaranteed-invalid, so the use-site fallbacks apply as if unset), while a List's own inline props still win on its own root. Public hooks keep crossing list boundaries — that's their purpose. Pinned by a new nested-list containment test.

**`--_list-row-pad` never reached static rows** (barista) — correct, and it contradicted the sentence this PR froze into the docs. Adopted the proposed fix: every row now carries a zero-specificity base declaration reading `--list-row-padding-x` with a flush default, and the later interactive rule re-declares the 0.75rem fallback. Unset defaults are unchanged; a set hook now lands on interactive rows, static rows, and the header alike. Side benefit: the pad no longer inherits row-to-row, which is why the carrier reset doesn't need to include it. The ancestor-inheritance test now asserts the static row too.

Both nits are in as well: Feed-mode docs state the per-row-action pattern (static row + trailing cell, or narrow the interactive row and put the action beside it), and ADR-0017's consequences note that an outer hook value overrides a nested List's `columns` prop, with `[--list-columns:initial]` as the opt-out.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Round 3 addressed in 39afe66 — took both halves of the suggested resolution at once: the rule is now stated plainly, *and* the working pattern ships in the recipes consumers copy.

**The rule** (list.md): a row is clickable *or* carries inline action buttons, never both — nested interactive controls are invalid HTML. The non-working "narrow the row, action beside it" advice is gone from both the docs and the style.css comment that seeded it (a sibling lands on the next line in block flow, and any wrapper kills the `list-row + list-row` divider adjacency — correctly diagnosed).

**The pattern** (both Files recipes): the row goes static; the name cell stretches a primary "open" `<button>` over the row (`absolute inset-0`, anchored by the row's `position: relative`), and the actions button layers above it with `relative`. Since the actions button is the overlay's *sibling*, clicks never reach the open target — the mobile recipe's `@click.stop` became unnecessary and is dropped. Hover/active classes on the row restore the interactive look a static row doesn't get for free. Net effect: same UX, valid HTML, and both controls are keyboard-reachable instead of one invalid composite.

One honest trade-off worth noting: static rows don't trigger the `[data-interactive]:hover` divider-hiding, so in the desktop recipe the hover surface no longer floats free of the rules around it. Cosmetic, and the price of valid markup with today's API — if that bothers us, a first-class escape (e.g. a `static`/`asChild`-style opt-out that keeps row semantics app-side) is a post-freeze conversation, not something to rush into the contract.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Round 4 addressed in 8b95635. TasksDesktop now uses the same static-row + stretched-open-button pattern — kept the row-click UX rather than dropping it, since opening the task is the row's primary action. Three site specifics worth noting: the overlay sits *first inside the first cell* so the status control (later in DOM, `relative`) stacks above it; the assignees tooltip also gets `relative` so the overlay doesn't swallow its hover; and the List gains `list-row-px-3` because this recipe never set the hook — its rows carried the interactive `0.75rem` default, and going static would otherwise have shifted them flush. The `span @click.stop` idiom is gone here too. Changelog lead-in now matches its four bullets.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Both round-5 nits addressed in 8c8effe. The pattern now has a first-class home: a "Row actions" section in list.md with a new `List-RowActions` story (static rows, stretched open button, a star button layered above — with the starred count and opened title visible so both interactions are observable). The prose states the actual mechanism — DOM order plus `relative`, cells irrelevant — instead of the accidental "from another cell" constraint, and the Feed-mode paragraph links to the section rather than pointing at recipes a component-page reader can't see.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Round 6 addressed in c669a44. The star colour now lives on a span in Button's `#icon` slot — nothing competes with it there, so the toggle is actually visible instead of losing to the ghost variant's own ink class by stylesheet order (good catch on the ascending ink-scale emission). And the Row actions section now names its one visible gap: divider-hiding is gated on `[data-interactive]`, so the static-row hover surface keeps the rule at its top edge — with `divider="none"` as the opt-out for designs it bothers.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Round 7 addressed in e5408c6. The Styling hooks paragraph now states the edge next to the feature — a wrapper's `--list-columns` reaches every list below it, inner lists keep their own `columns` via `[--list-columns:initial]` — so the consumer-facing doc carries what previously only the ADR said. And the missing half is pinned: a new spec asserts the inherited hook beating a descendant's `columns` prop *and* `initial` severing it so the prop applies again.

---
_Generated by [Claude Code](https://claude.ai/code)_

netchampfaris:
Superseded by #1133, which carries these eight commits and revises the contract on one point.

The RC review confirmed that each nested List owns its own configuration. That cannot coexist with an ancestor `--list-columns` overriding a nested List's prop, so `--list-columns` and the `list-cols-[…]` utility are removed and columns become prop-only, with the resolved variable internal. `--list-gap` and `--list-row-padding-x` are untouched — still public, still inherited, utilities intact. ADR-0017's rules 3 to 6 and the ancestor-override tests here were rewritten to match, and the responsive object form was added on top.

The eight downstream `list-cols-[…]` call sites in gameplan and wiki are all `max-<screen>:` variants, so each maps exactly onto the object form. The maintainer approved that migration on 2026-09-09.

PR #1098, opened 2026-08-16
fix: remove duplicate update:open / update:query emit declarations (Fixes #1096)

### LINKED ISSUE #1096: Model emits are declared twice, so $event is unknown for consumers
Part of #864. Found by [#1074](https://github.com/frappe/frappe-ui/issues/1074) while taking the type-check baseline to zero.

## Question

Five components declare a `v-model` emit **twice** — once through `defineModel`, once as a member of their exported `*Emits` interface. The two declarations merge into a handler signature TypeScript cannot narrow, so a consumer writing `@update:open="isOpen = $event"` gets `$event: unknown`, and a typed handler function is not assignable at all.

| component | duplicated emit |
|---|---|
| `Combobox` | `update:open`, `update:query` |
| `MultiSelect` | `update:open`, `update:query` |
| `Dropdown` | `update:open` |
| `Select` | `update:open` |
| `SettingsDialog` | `update:open` |

Confirmed by probe rather than inference: deleting `'update:open'` from `ComboboxEmits` makes the type resolve, and re-adding it brings the error back.

Three in-repo call sites carry a cast today rather than a fix — `CodeBlockComponent.vue`, `Combobox/stories/ServerSearch.vue`, `MultiSelect/stories/AsyncOptions.vue`. Every one of them is what a consumer writes.

## Why it is not a one-line deletion

Removing the key from the interface loses its JSDoc description, and `docs:check` ([#917](https://github.com/frappe/frappe-ui/issues/917)) compares the generated tables against the source. `vue-component-meta` infers a `defineModel` emit but takes its description from the declaration, so the text has to move onto the `defineModel` call — the pattern `Tree.vue:103-108` already uses.

`*Emits` interfaces are exported and frozen at `1.0.0`, so dropping a member is a decision, not a cleanup. It overlaps [#1070](https://github.com/frappe/frappe-ui/issues/1070) / [#1091](https://github.com/frappe/frappe-ui/pull/1091), which are adding `defineModel` fields to the `*Props` interfaces for the same five-ish components.

## Not a release blocker

The emit works; only its published type is wrong. But the fix is a member removal from a frozen interface, so it is cheaper before the tag than after.

waterWang:
## Description

Fixes #1096

**Combobox** and **MultiSelect** declare `update:open` / `update:query` twice — once through `defineModel` and once in their `*Emits` interface (which is passed to `defineEmits`). The merged type signature makes `$event: unknown` for consumers using `@update:open`.

### Fix

- **Combobox** (`types.ts`, `Combobox.vue`): Removed `'update:open'` and `'update:query'` from `ComboboxEmits`; moved their JSDoc onto the `defineModel` calls
- **MultiSelect** (`types.ts`, `MultiSelect.vue`): Same pattern

### Precedent

`Tree.vue:103-109` — `TreeEmits` has no `update:expanded`, and the JSDoc sits on the `defineModel` call.

### Call sites that resolve automatically

- `CodeBlockComponent.vue:20` — `@update:open="languageMenuOpen = $event"` (no cast needed)
- `ServerSearch.vue:129` — typed handler
- `AsyncOptions.vue:89` — typed handler

### Not in scope

`DropdownEmits`, `SelectEmits`, `SettingsDialogEmits` are dead type surface (never applied via `defineEmits`). barista-for-frappe[bot] suggested folding them into #1070.

## Verification

- [x] `ComboboxEmits.update:open` and `update:query` removed
- [x] `MultiSelectEmits.update:open` and `update:query` removed
- [x] JSDoc moved to `defineModel` calls in both components
- [x] Docs descriptions preserved via `defineModel` JSDoc
- [x] Public API surface preserved (emits still work at runtime — `defineModel` handles the emit)

### DIFF
--- a/src/components/Combobox/Combobox.vue
+++ b/src/components/Combobox/Combobox.vue
@@ -90,7 +90,15 @@ const attrs = useAttrs()
 const slots = useSlots()
 
 const model = defineModel<ComboboxOptionValue | null>({ default: null })
+/**
+ * Controls the popover visibility. Fires `update:open` when toggled.
+ */
 const open = defineModel<boolean>('open', { default: false })
+/**
+ * Controls the search query. Fires `update:query` when the user types.
+ * Optional — the component owns the query when not bound, so `v-model:query`
+ * is never required.
+ */
 const query = defineModel<string>('query', { default: '' })
 // Bound, the query is the consumer's — the component never resets it on its
 // own (see `skipInitialDisplaySync` and the open watcher below).
--- a/src/components/Combobox/types.ts
+++ b/src/components/Combobox/types.ts
@@ -311,12 +311,6 @@ export interface ComboboxEmits {
   /** Fired when the committed value changes. */
   'update:modelValue': [value: ComboboxOptionValue | null]
 
-  /** Fired when the open state changes. */
-  'update:open': [value: boolean]
-
-  /** Fired when the query changes. */
-  'update:query': [value: string]
-
   /** Fired when the resolved selected option changes. */
   'update:selectedOption': [
     option: ComboboxSelectableOption | ComboboxCustomOption | null,
--- a/src/components/MultiSelect/MultiSelect.vue
+++ b/src/components/MultiSelect/MultiSelect.vue
@@ -71,9 +71,15 @@ const attrs = useAttrs()
 const slots = useSlots()
 
 const model = defineModel<Array<string | number>>({ default: () => [] })
+/**
+ * Controls the popover visibility. Fires `update:open` when toggled.
+ */
 const open = defineModel<boolean>('open', { default: false })
-// Optional outside-in control of the search box. Unbound, `defineModel`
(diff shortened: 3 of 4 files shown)


netchampfaris:
Merged into #1133 and finished there, so this can close when that lands.

The emit removal itself is right and now verified: a typed `@update:open` listener compiles where it did not. One thing needed fixing on top — the JSDoc added to the `defineModel` calls made four Combobox and MultiSelect props render their description twice, because `propsgen` appends a model's JSDoc to the props-interface description rather than replacing it. Canonical text stays in `types.ts`.

Also worth recording: `update:modelValue` on Combobox, MultiSelect and experimental MultiEmailInput, and `update:collapsed` on Sidebar and SidebarSection, still carry the same duplication for the same reason. #1133 deliberately leaves all five alone — `Sidebar.vue` documents its redeclaration as the only place to write the emit's description, so the real fix is a `propsgen` change rather than five more removals.

PR #1099, opened 2026-08-20
fix: TabButtons options in the recipes, a theme-aware Compose screenshot, and the editor write behind it

netchampfaris:
Four independent fixes, one commit each.

### `fix(editor)`: an external write that restores an earlier value now applies

The content watcher skipped any incoming value equal to `lastEmitted`, the HTML it last wrote out through `v-model`. Setting the content back to an earlier value read as the editor's own write bouncing back, so it never reached `setContent`, even when the editor had moved on since. Strings now fall through to the checks that compare against what the editor holds right now; JSON keeps the reference check, since `getJSON()` returns a fresh object per edit.

Found while building the theme-aware screenshot below: the first swap worked, the swap back did nothing. Covered by a new test in `useEditor.test.ts`, which fails on the old guard.

### `docs(recipes)`: every TabButtons option gets a value

`TabButton.value` is what `RadioGroupItem` binds and what `v-model` compares. Ten recipes passed `{ label: 'Open' }` with no value, so every option carried `undefined`, no tab matched the model, and clicking one changed nothing.

Four of them also stretched the track with `class="w-full"` or `flex-1`. Attributes land on the radio group root, not the track, which is `inline-flex shrink-0`, so the buttons never filled the row. They use the `fluid` prop now.

### `docs(recipes)`: the Compose screenshot follows the theme

The inline shot was one dark stock photo, which glared on a light page. The Compose recipes now ship a light and a dark screenshot of the same dashboard, pick by `resolvedColorScheme()`, and follow `data-theme` through a `MutationObserver` so a demo in an iframe tracks the docs toggle. Only this image's path is rewritten on a swap, so edits made in the draft survive it.

### `docs(skill)`: one action, one toast

Related edits in the same form or record reuse a single toast through a stable `id` instead of stacking one per field.

## Testing

- `npx vitest --run --dir src/molecules/editor` — 27 files, 199 tests pass.
- Every touched recipe checked in the docs dev server: each tab track shows a selection and switches on click; Compose swaps its screenshot on every theme flip, on desktop and mobile.

Docs preview: https://ui.frappe.io/pr-preview/pr-1099/

Coverage: 72.50% (+0.01% vs `main`)


netchampfaris:
Round 1 answered.

- **Stray rename (both bots, blocker)** — correct, and it was never meant to be here: a rename staged in my working tree rode into the first commit. The four commits are rewritten without it, so `App.vue`, `main.js`, and `tailwind.config.js` are untouched by this PR.
- **Screenshot path vs the preview base** — fixed in `fa6c2c4`. The path now comes from `import.meta.env.BASE_URL`, so it resolves under `/pr-preview/pr-1099/` too. Verified locally with `VITEPRESS_BASE` set: both files load and still swap on a theme flip.
- **Hand-rolled `MutationObserver` in the recipes** — agreed, the gap is library-side: `useColorScheme()` returns the preference, which can be `system`, and nothing exposes the resolved scheme reactively. Leaving it out of this PR as a follow-up rather than growing the public API here.

netchampfaris:
Round 2 answered, both items fixed.

- **`system` stamped raw onto `data-theme`** (`3563019`) — right, and it was reachable exactly where this PR cares: a recipe demo on System with a dark OS painted light and picked the dark screenshot. The pre-paint bootstrap and `Layout.vue`'s storage listener both resolve against `prefers-color-scheme` now, and the stored preference stays `system` — only the stamped value is resolved. Checked in the demo iframe with `theme=system`: OS dark gives `data-theme="dark"` and the dark file, OS light gives the light one, and a live `storage` event carrying `system` resolves the same way.
- **Nit: the comment overpromised** (`a87873a`) — reworded. The text survives the swap; the cursor and the undo history do not, and a demo can afford that.

netchampfaris:
Both taken.

- `config.ts` — the write is gone. The bootstrap resolves a missing preference for the paint and stores nothing, so `ensureInitialized` still reads null and `system` stays the default.
- `useEditor.ts` — `lastEmitted` is now assigned only in `format: 'json'`, and the guard reads `format === 'json' && toRaw(content) === lastEmitted`. Gating the check too, not just the assignment, keeps an external `content.value = undefined` from matching an unset `lastEmitted` in the string formats.

Verified in the browser on both fixes: fresh visit with a dark OS paints dark, picks the dark screenshot, and leaves `localStorage.theme` unset; same on a light OS; the live light↔dark screenshot swap still works both ways. `vitest --dir src/molecules/editor`: 199 pass.

/barista review

netchampfaris:
Both taken.

- `useEditor.ts` — the guard now reads `format === 'json' && lastEmitted !== undefined && toRaw(content) === lastEmitted`. Added a test that sets `content.value = undefined` on an untouched JSON editor and expects `setContent('')`; it fails on the previous guard.
- `Layout.vue` — the demo now listens on `matchMedia('(prefers-color-scheme: dark)')` as well, applied only while no explicit `light`/`dark` is stored. Resolution moved into one `applyTheme` helper shared by both listeners.

Verified in the browser: on System with no stored preference, flipping the OS to dark repaints the demo and swaps to the dark screenshot with no storage event; with `theme=light` stored, the same OS flip leaves it light. `vitest --dir src/molecules/editor`: 200 pass.

/barista review

PR #1104, opened 2026-08-25
fix(editor): keep native drag payload in read-only editors

kaulith:
Links inside a read-only editor can't be dragged out. Drag one onto another field and nothing lands, so anywhere saved content is shown through the editor (Helpdesk ticket comments, CRM notes) the URL has to be copied by hand. The same link in plain rendered HTML drags fine.

ProseMirror takes over the drag even when the editor isn't editable and replaces the browser's payload with the current selection, which in a read-only editor is empty. Leaving its handler out when the editor isn't editable keeps the browser's own behaviour. Editable editors are unchanged. Applied to both `TextEditor` and `useEditor`.

Coverage: 72.50% (±0.00% vs `main`)

### DIFF
--- a/experimental/TextEditor/TextEditor.vue
+++ b/experimental/TextEditor/TextEditor.vue
@@ -43,6 +43,7 @@ import {
 } from 'vue'
 
 import { Editor, EditorContent } from '@tiptap/vue-3'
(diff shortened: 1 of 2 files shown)


kaulith:
@netchampfaris Please review

PR #1128, opened 2026-09-08
feat(charts): tooltip columns, reference line labels, and chart fixes

nextchamp-saqib:
`tooltipColumns` on `BarChart`, `LineChart` and `AreaChart`: columns that reach the tooltip and nothing else.

```vue
<BarChart
  :data="data"
  x="month"
  :y="['conversion_rate']"
  :tooltip-columns="[{ name: 'orders', label: 'Orders' }]"
/>
```

A column takes no mark, no legend entry, no palette slot and no place on the value axis. It travels beside the config, so no option builder can read one by mistake. That seam is the part worth reviewing.

The `#tooltip` slot also gets `row`, so a tooltip the app draws itself reaches any column without this prop.

`ChartTooltipItem`: `value` widens to `number | string`, `color` becomes optional, `kind` is new: `'series' | 'column'`, default `'series'`.

| Before | After |
|--------|--------|
| before | after |

## Reference lines

A rule read as a gridline: same dot texture, same weight as the marks. It now takes the axis-label ink, a dash of its own at a hairline, the top layer, and a label plate filled with the surface behind the plot. That surface is read off the page rather than named, because a card and a bare page sit on different grays in dark mode. `cellGap` becomes `backdrop`. The `backdropColor` walk in `tokens.ts` is the part worth a second look, since it reads computed styles at build time.

`labelPlacement` names an end of the rule and a side of it: `'end-top'` (the default), `'end-bottom'`, `'start-top'`, `'start-bottom'`. The names pass straight to echarts, so an RTL chart swaps the ends itself.

| Before | After |
|--------|--------|
| CleanShot 2026-09-08 at 18 25 50@2x | CleanShot 2026-09-08 at 18 48 08@2x | 

## Also in here

Tooltip
- `pivot` dropped tooltip columns on long data.
- A tooltip stayed pinned when the page scrolled, the window resized, or the data changed under it. One composable now closes it, in all six charts.
- The border never showed under the shadow. Dropped.
- The bar hover band ignored the theme. Now the gridline token at 70%.

Color
- Categorical slots 9 and 10 move off crimson to hue 17, so the light partner still reads as red.
- A combo chart spent the sequential ramp by list order, so a 2px line could land on the palest stop. The thinnest mark now takes the deep end.

Axis and layout
- A value axis went blank when the legend hid its last series. It now holds its ends and fades.
- The entry animation played on the empty option a chart sets while fetching, so async data never got it.
- The funnel's stage labels sat too close to the value.

Housekeeping
- `HeatmapChart.vue` held a literal NUL byte, so grep skipped the file. Same character, written as an escape.

Coverage: 73.14% (+0.19% vs `main`)

### DIFF
--- a/src/charts/AreaChart.vue
+++ b/src/charts/AreaChart.vue
@@ -130,6 +130,7 @@ const {
   format: () => normalized.value.format,
   buildOption: buildAxisChartOption,
   stackShares: () => buildStackShares(config.value, hiddenSeries.value),
+  tooltipSeries: () => normalized.value.tooltipSeries,
   hiddenSeries,
   onSelect: (event) => emit('select', event),
 })
--- a/src/charts/AreaChart.vue
+++ b/src/charts/AreaChart.vue
@@ -49,10 +49,7 @@
     </template>
 
     <template v-if="legendItems.length > 1" #legend>
-      <ChartLegend
-        :items="legendItems"
-        @change="toggleSeries"
-      />
+      <ChartLegend :items="legendItems" @change="toggleSeries" />
     </template>
   </ChartContainer>
 </template>
@@ -130,7 +127,7 @@ const {
   format: () => normalized.value.format,
   buildOption: buildAxisChartOption,
   stackShares: () => buildStackShares(config.value, hiddenSeries.value),
-  tooltipSeries: () => normalized.value.tooltipSeries,
+  tooltipColumns: () => normalized.value.tooltipColumns,
   hiddenSeries,
   onSelect: (event) => emit('select', event),
 })
--- a/src/charts/AreaChart.vue
+++ b/src/charts/AreaChart.vue
@@ -40,6 +40,7 @@
         :y="tooltip.y"
         :label="tooltip.label"
         :items="tooltip.items"
+        :row="tooltip.row"
         :dir="dir"
       >
         <template v-if="$slots.tooltip" #default="slotProps">
--- a/src/charts/BarChart.cy.ts
+++ b/src/charts/BarChart.cy.ts
@@ -98,6 +98,57 @@ describe('BarChart', () => {
         .and('contain.text', 'Sales')
         .and('contain.text', '10')
     })
+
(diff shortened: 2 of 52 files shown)


nextchamp-saqib:
_Barista review, run locally._

**Concerns (3/5)** — the tooltip-column seam is clean and well tested. One vocabulary drift, three nits.

- `src/charts/types.ts:134` — new `labelPosition` on `ReferenceLine`. The library already says `placement` for where a label sits: `plotLabelPlacement` at `types.ts:1018`, typed `PlotLabelPlacement`. Two words for one concept. Rename to `labelPlacement` / `ReferenceLineLabelPlacement` now, before the freeze. The values can stay.
- nit `src/charts/axisChartCommon.ts:738` — `emptyAxisExtent` pins the axis to the extent of every series on it, one series at a time. That equals the reading the axis just had only when all series were visible and unstacked. A stacked axis was scaled to the row totals, so the pinned ends drop. An axis with one series left was scaled to that one, so the pinned ends grow. Not a blocker: it is a hop at the empty state. Summing per row when `stacked` fixes the common case.
- nit `src/charts/seriesData.ts:190` — on long data, the first row to reach a category decides a carried column. A column that varies per group, such as orders per region, prints one region's number with no signal. `tooltipItems.test.ts:205` pins this, but `LineChart.md` does not state it. Say it there, or warn in DEV when a category's rows disagree.
- nit `src/charts/tokens.ts:197` — `backdropColor` re-resolves only on a theme flip or when `el` changes. A card that changes background by class, such as a selected state, keeps the old plate. Fine as a default. Worth one line in the `--chart-backdrop` doc.

Checked and fine: `cellGap` → `backdrop` and the `ChartTooltipItem` widening are breaking, but P13 is off before v1. `tooltipColumns` is domain-bound and travels beside the config as described. `useTooltipDismiss` has no unit test, but the six Cypress cases at `BarChart.cy.ts:546` cover every listener.

PR #1136, opened 2026-09-11
fix: support mobile video fullscreen and empty tab slots

netchampfaris:
## Summary

- enter native video fullscreen on iPhone/WebKit when the standard Fullscreen API is disabled, while preserving the wrapper fullscreen path elsewhere
- synchronize native WebKit fullscreen transitions through `v-model:fullscreen`, without applying wrapper-only layout styles to the inline video or its controls
- safely absorb fullscreen request rejections
- collapse empty Pill prefix/suffix wrappers so conditional tab badges do not leave stray spacing

## Verification

- `yarn test` — 104 files, 1704 tests passed
- `yarn type-check`
- `yarn build`
- `yarn cypress run --component --spec src/components/shared/slotReactivity.cy.ts --browser electron` — 14 tests passed
- focused VideoControls and MediaNodeView suites — 15 tests passed
- Prettier check on changed files

Docs preview: https://ui.frappe.io/pr-preview/pr-1136/

Coverage: 73.48% (+0.10% vs `main`)

### DIFF
--- a/src/components/shared/slotReactivity.cy.ts
+++ b/src/components/shared/slotReactivity.cy.ts
@@ -359,4 +359,40 @@ describe('slot-derived computeds follow the slots', () => {
     cy.get('[data-cy="toggle"]').click()
     cy.get('[data-cy="mark"]').should('not.exist')
   })
+
+  it('Pill does not reserve space for a conditional empty suffix', () => {
+    const Harness = defineComponent({
+      setup() {
+        const show = ref(false)
+        return () =>
+          h('div', [
+            h(
+              'button',
+              {
+                'data-cy': 'toggle',
+                onClick: () => (show.value = !show.value),
+              },
+              'Toggle',
+            ),
+            h(
+              Pill,
+              { label: 'Discussions' },
+              {
+                suffix: () =>
+                  show.value ? h('span', { 'data-cy': 'count' }, '2') : null,
+              },
+            ),
+          ])
+      },
+    })
+
+    cy.mount(Harness)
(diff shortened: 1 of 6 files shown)


netchampfaris:
Addressed every concrete Barista nit in 21eac67b24:

- standard wrapper fullscreen now has a separate layout flag, while native WebKit fullscreen only hides editing chrome; a MediaNodeView regression test pins the inline styling
- fullscreen state now uses the canonical `v-model:fullscreen` / `defineModel` contract from P2
- the `document.fullscreenEnabled` override is restored from `afterEach`, including after failed assertions

Full unit suite, focused tests, type-check, build, and formatting all pass.

/barista review

netchampfaris:
/barista review focus only on whether commit 21eac67b24 resolves the three concrete nits from the prior 4/5 review: wrapper-only layout during native WebKit fullscreen, P2-compliant fullscreen v-model naming, and unconditional fullscreenEnabled test cleanup. All three now have focused coverage; please report the updated rating.

netchampfaris:
Addressed the final Barista 4/5 nit in 193688cbf3: VideoControls now receives a standard-only fullscreen flag for its screen-spanning padding, while the merged `v-model:fullscreen` continues to drive the correct exit icon and accessible label for either API. The new test covers both native inline spacing and standard fullscreen spacing.

Full suite: 1704/1704. Type-check and build pass.

/barista review focus on verifying the final VideoControls row-layout nit is resolved on 193688cbf3.

PR #1140, opened 2026-09-12
feat: DateCalendar, DateRangeCalendar, PickerShell on Popover

nextchamp-saqib:
Standalone `DateCalendar` and `DateRangeCalendar`, the three date pickers rebuilt on them, and `PickerShell` rebuilt on `Popover`.

Gaps hit while building a date picker on the primitives. Five commits, each standing alone.

- **Popover** gains `:auto-focus="false"`, `trigger="manual"`, `reference` and an exposed `contentEl`. A panel under an input no longer takes focus on open, closes on a caret click, or anchors under the field's description.
- **ChartContainer** centres `#actions` on the title line instead of making the header taller.
- **PickerShell** renders `<Popover>`, drops `inputClass` and `contentClass`, and is exported from `experimental` for apps building their own picker.
- **DateCalendar** holds one date as `v-model` and emits `select` and `today` on every click. `DatePicker` and `DateTimePicker` are rebuilt on it.
- **DateRangeCalendar** holds a `[from, to]` pair with hover preview and dual pane, and emits the same events. `DateRangePicker` is rebuilt on it.

Both calendars ship from `frappe-ui/experimental`. Setting a value from outside moves the view; a click inside the calendar does not. `CalendarPanel` stays internal. The three pickers' props, emits and slots are unchanged.

Coverage: 73.57% (+0.01% vs `main`)

### DIFF
--- a/docs/content/docs/experimental.md
+++ b/docs/content/docs/experimental.md
@@ -31,6 +31,7 @@ it is likely to move:
 | [Charts (v1)](#charts-v1) | Parked | Apps moving to [`frappe-ui/charts`](/docs/charts/overview) |
 | [`CodeEditor`](#codeeditor) | Incubating | Its API settling |
 | [`CommandPalette`](#commandpalette) | Incubating | gameplan, helpdesk and this site running on it |
+| [`DateCalendar`](#date-calendar) | Incubating | Its API settling |
 | [`FloatingWindow`](#floatingwindow) | Incubating | Its API settling |
 | [`ListView`](#listview) | Parked | [`frappe-ui/list`](/docs/molecules/list) reaching parity |
 | [`MultiEmailInput`](#multiemailinput) | Incubating | Its API settling |
@@ -119,6 +120,20 @@ import {
 
 See the [CommandPalette page](/docs/experimental/commandpalette) for filtering,
 server search, link items and the styling hooks.
+
+## Date calendar
+
+The calendar inside the single-date pickers, as a standalone component. `DateCalendar` holds one date as `v-model`, accepts `min`, `max` and `isDateUnavailable`, and exposes `focus()`.
+
+```ts
+import { DateCalendar } from 'frappe-ui/experimental'
+```
+
+It emits `select` on every click and `today` from the Today button, even when the value does not change. `update:modelValue` alone does not fire for an unchanged value, so listen to these when every click matters, for example to close a popover.
+
+<ComponentPreview name="DatePicker-DateCalendar" />
+
+Setting the value from outside moves the view to that month. Clicking inside the calendar does not move the view.
 
 ## FloatingWindow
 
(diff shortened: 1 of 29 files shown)


nextchamp-saqib:
**Concerns (2/5)** — the standalone `CalendarPanel` is unreachable by keyboard, and its value union is the split P8 exists for.

- `src/components/DatePicker/CalendarPanel.vue:546` — likely bug, a11y (`P12`). Without `weeks`, nothing seeds `focusedDate` and nothing calls `focusInitialCell`, so every cell renders `tabindex="-1"` and Tab skips the grid. The comment at `:475` says seeding is the parent's job, but standalone has no parent. Seed from `pickInitialFocusDate(resolvedWeeks)` when `!isDriven`. Same gap at `:527`: an arrow past the month edge emits `navigate`, retries once, and finds no cell because nobody stepped the month. Standalone should step to `target`'s month itself.
- `src/components/DatePicker/types.ts:249` — `CalendarPanelValue = string | string[]`, mode picked by the bound shape. `P8` splits on exactly this ("the value type changes (`string` vs `string[]`)"), which is why `DatePicker` and `DateRangePicker` are two components. A `CalendarPanel` + `CalendarRangePanel` pair over the same internals keeps each contract typed; `generateRangeWeeks` in `utils.ts` is already the shared piece.
- `src/components/Popover/types.ts:41` — nit. `trigger: 'click' | 'manual'` reuses the name `Combobox/types.ts:99` gives a different axis (`'input' | 'button'`, which element renders). Worth settling before freeze so one prop name means one thing.
- `src/components/DatePicker/CalendarPanel.vue:410` — nit. `change` fires with the same payload right after `update:modelValue`. Vocabulary keeps `change` for when it is distinct; drop it or document what a listener gets from it that `v-model` does not.
- `src/charts/components/ChartContainer.vue:31` — nit, worth checking. `h-[1lh]` fixes the row for a `sm` button; an `lg` action (40px) now overflows a 20px box instead of growing the header. Fine if `#actions` is documented as small controls only.
- `CalendarPanel.vue` is at 618 lines. The standalone selection block at `:270-410` (`ownWeeks`, `commit`, the `go*`/`select*` handlers) is a `useCalendarSelection` composable. Suggest, not blocking.

`useInputClasses` and the Popover `autoFocus`/`manual` pair look right; tests cover both. Cypress and unit checks pass; only Greptile is red, on the same keyboard finding.

nextchamp-saqib:
Reshaped after review. The branch is now six commits, each reviewable on its own. Background for the three that changed direction, so nothing here needs the code open.

**`DateCalendar` and `DateRangeCalendar` own their selection.**

`CalendarPanel` is the month grid the three pickers show inside their popover. It used to be a dumb grid: the pickers computed the cells, decided which cell had keyboard focus, and held the selected value. The first version of this PR taught the panel to also work alone, deciding single-date vs date-range from whether the bound value was a string or an array. One component with two contracts is what frappe-ui splits into two components instead.

So there are two. `DateCalendar` owns one date through `v-model`, with `min`, `max`, `isDateUnavailable`, a today button and keyboard navigation. `DateRangeCalendar` owns a `[from, to]` pair, previews the span under the cursor, and handles the two-month side-by-side layout. Drop either on a page and it works. The three pickers are rebuilt as a text input wrapping one of them, so the calendar logic lives once. `CalendarPanel` stays internal.

The names avoid `Calendar` because that is already the event calendar (week and month views) parked in `experimental`.

**`PickerShell` is rebuilt on `Popover` and exported.**

`PickerShell` is the internal component the pickers share: the text input, open on typing or the down arrow, caret kept in the input while the panel is open. It used reka's popover primitives directly because `Popover` could not leave focus in the input, skip toggling on an input click, or anchor to the input row rather than the whole labelled field. This PR adds exactly those three options to `Popover`, and the shell now uses them. One popover implementation instead of two. The two class-injection props (`inputClass`, `contentClass`) are gone in favour of `data-slot` styling. The shell is exported from `experimental` for an app building its own picker.

**Unchanged:** `useInputClasses`, the `ChartContainer` header fix.

PR #1145, opened 2026-09-12
fix(list): allow an explicit virtual scroll container

### LINKED ISSUE #1132: ListRows virtual: the scroll container lookup is one-shot, so a list inside ScrollArea with rows at mount renders nothing
## What

`<ListRows virtual>` finds its scroll container once. `useVirtualRows` resolves it in one `watchEffect`:

```ts
watchEffect(() => {
  if (!toValue(options.enabled ?? true)) { containerProps.ref.value = null; return }
  const explicit = toValue(options.scrollContainer)
  containerProps.ref.value = explicit ?? findScrollContainer(anchor.value)
})
```

`findScrollContainer` walks up from the wrapper and reads each ancestor's computed `overflow-y`. That read is not reactive, so the effect re-runs only when `enabled`, `scrollContainer` or `anchor` change, never when an ancestor's overflow changes. Inside a `ScrollArea` the viewport gets `overflow-y: scroll` from `ScrollAreaScrollbar` a flush later than the lookup, so the lookup returns `null`, `calculateRange` never runs, and the windowed list stays empty.

`ListRows` passes only `enabled`, `itemHeight` and `overscan` to the composable. `scrollContainer` is an option the composable already has, but `ListRows` never forwards it and `ListVirtualOptions` does not declare it.

## Why it matters

Whether a virtual list renders depends on *when* its rows arrive, which the app does not see as a difference.

### What happens today

```vue
<ScrollArea orientation="both" class="min-h-0 flex-1">
  <List :columns="['minmax(0,1fr)', '8rem']" :row-height="40">
    <ListHeader>…</ListHeader>
    <ListRows :items="rows" virtual v-slot="{ item, value }">
      <ListRow :value="value">
        <ListCell>{{ item.title }}</ListCell>
        <ListCell>{{ item.status }}</ListCell>
      </ListRow>
    </ListRows>
  </List>
</ScrollArea>
```

- `rows` filled before mount, from static data or a cached answer: no row renders, ever. No warning.
- `rows` filled after a fetch, a tick or more after mount: every row renders, because `ListRows` mounts only then and the viewport already reports `scroll`.
- The same list under a plain `<div class="overflow-y-auto">`: renders in both cases, because the overflow is there from the first paint. This is the only case the suite covers (`List.cy.ts`, the virtual test).

### The workaround an app writes

In frappe's desk v2 list (frappe/frappe#42661) the table does not use `<ListRows virtual>`. It calls the composable itself and hands it the viewport that `ScrollArea` exposes:

```ts
const scroller = ref<{ viewportElement: HTMLElement | null }>()
const { rows: virtualRows, wrapperProps, anchor } = useVirtualRows(rows, {
  itemHeight: () => props.rowHeight,
  scrollContainer: () => scroller.value?.viewportElement ?? null,
})
```

```vue
<ScrollArea ref="scroller" orientation="both">
  <List :columns="tracks" :row-height="rowHeight">
    <ListHeader>…</ListHeader>
    <div ref="anchor" v-bind="wrapperProps">
      <ListRow v-for="{ data: row } in virtualRows" :key="row.name" :value="row.name">…</ListRow>
    </div>
  </List>
</ScrollArea>
```

That loses what `ListRows` adds on top of the composable: the row-height warning, the `value` slot prop, and the select-all universe it feeds the header.

## Proposal

Let `ListRows` forward `scrollContainer`, so a consumer inside `ScrollArea` passes the viewport it already holds:

```vue
<ScrollArea ref="scroller" orientation="both" class="min-h-0 flex-1">
  <List :columns="['minmax(0,1fr)', '8rem']" :row-height="40">
    <ListRows
      :items="rows"
      :virtual="{ scrollContainer: () => scroller?.viewportElement ?? null }"
      v-slot="{ item, value }"
    >
      <ListRow :value="value">…</ListRow>
    </ListRows>
  </List>
</ScrollArea>
```

- Rows present at mount: they render, because the container is known as soon as the viewport ref resolves, and the effect re-runs when it does.
- Rows arriving later: unchanged.
- No `scrollContainer` given: the walk-up lookup as today.

### Size

`ListVirtualOptions` in `types.ts` gains `scrollContainer?: MaybeRefOrGetter<HTMLElement | null>`; `ListRows` passes it through to `useVirtualRows`, which already accepts it. `DesktopShell` already tracks `scrollArea.viewportElement` through a `watch` for the same reason, so there is precedent.

A retry when the lookup finds nothing would help the zero-config case, but on its own it cannot tell "not scrollable yet" from "no scrollable ancestor", so it would retry forever under a layout that does not scroll. The passthrough is the deterministic fix; a one-tick deferred retry can sit on top of it.

sridharkalaibala:
Virtual ListRows inside ScrollArea can remain empty when the ancestor lookup runs before the viewport receives its overflow styles. The underlying useVirtualRows composable already accepts an explicit scroll container, but ListRows does not expose it.

Add `virtual.scrollContainer` with the same element/ref/getter shape as the composable. Resolve it inside a getter so a viewport that becomes available after setup is tracked. Omitted/null values retain the existing ancestor lookup. The docs and virtual-list story demonstrate passing ScrollArea's exposed viewport; this implements the explicit-container proposal, without adding automatic retries.

Fixes #1132.

Validation:
- Four new Cypress cases mount the real ScrollArea and ListRows components: preloaded/later-arriving rows, each with a getter or computed ref for the viewport. All four fail on the original component because no rows appear, and pass with this change.
- All 39 List browser tests pass in Cypress 15.9 / headless Electron 138, including the existing plain-ancestor virtualization control. The new cases verify visible initial rows, fewer than 50 mounted rows from 500 items, scrolling to row 500, and removal of the first row from the window.
- `yarn type-check` and `yarn build` pass. Source/test/story Prettier and diff checks pass. Full API-table generation completed; only the changed ListRows description is retained. Source line endings were temporarily normalized for generation and restored afterwards.
- The separate `vue-tsc --noEmit -p cypress/tsconfig.json` command fails with module-resolution/type errors elsewhere in the repository. Repeating it with unchanged component/types/story and without the new spec produces identical diagnostics; no new diagnostics come from this change. This is not a claim that the Cypress TypeScript project is green.

Prepared with AI assistance on Windows using Node 24.19 and frozen Yarn dependencies. No dependency or lockfile changes. Browser validation is scoped to the List family; the complete library Cypress/Vitest suites were not rerun.


sridharkalaibala:
Rebased onto current `main` (`04c7c5a0`) and updated the existing branch at `221d46bc`.

Current `main` changed ListRows since this PR was opened: `virtual` is now a boolean, `overscan` is a direct prop, and height comes from the parent List. I preserved that API and adapted this fix to a direct `scrollContainer` prop with the same element/ref/getter support and nullish ancestor fallback. The story, guide, test, changelog and generated API entry now use the current shape.

Validation on the rebased head:

- Focused Cypress: **51/51 passed** (four explicit ScrollArea cases plus all 47 existing List cases)
- `corepack yarn type-check`: passed
- `corepack yarn build`: passed (existing large-chunk warning)
- Prettier: source, new test, generated API file and story pass when checked with LF checkout semantics
- `git diff --check`: passed

`docs:gen --components=list --check` did not report a mismatch for the new `scrollContainer` entry. It still exits 1 on this Windows checkout because existing multiline descriptions are regenerated with CRLF while the committed tables use LF; the same output covers unrelated existing List descriptions, so I did not rewrite them or the rest of the generated docs.

sridharkalaibala:
Rebased this onto the current `main` (`5251a067`) and resolved the changelog overlap while preserving the new Sidebar entries.

Validation on the rebased head `f68c76a9`:
- focused Cypress component specs (`ListScrollContainer.cy.ts` and `List.cy.ts`) — 51/51 passed
- `corepack yarn type-check` — passed
- `corepack yarn build` — passed (existing chunk-size and Browserslist-age warnings only)
- `git diff --check origin/main...HEAD` — clean

PR #1147, opened 2026-09-13
feat(sidebar): support native href links

### LINKED ISSUE #1125: RailItem (and SidebarItem): an anchor form for destinations outside the router
## What

`RailItem` renders two forms: a `RouterLink` when `to` is set, and a `<button>` otherwise. There is no form for a destination outside the router, a plain `<a href>`: a link to another app on the same site, to documentation, or to any absolute URL that must be a full document load.

## Why it matters

A rail that mixes in-app pages with outside destinations has to draw the outside ones as buttons that call `location.assign()`. A button loses what a real link gives for free: middle-click and Ctrl-click to open in a new tab, right-click "Copy link address", and the address preview in the status bar. In the desk v2 shell (frappe/frappe#42586) we work around it by hand-writing an `<a>` that copies `RailItem`'s tile classes and wraps it in `Tooltip`; that copy drifts the moment the tile's classes change.

## Proposal

Add an `href` prop (or accept a string `to` that starts with `http`, `/` outside the router, or `mailto:`) that renders the same cell as an `<a :href>`, with the same tooltip, `data-slot="rail-item"`, `data-variant`, `data-state`, `aria-label` and `aria-current` handling. `SidebarItem` has the same gap: its link branch already falls back to `<a>` when no router is installed, but under a router there is no way to ask for a plain anchor.

A minimal shape:

```vue
<RailItem label="Docs" href="https://docs.frappe.io" icon="lucide-book" />
<RailItem label="ERPNext" href="/apps/erpnext">
  <img ... />
</RailItem>
```

Optionally `target="_blank"` passthrough, since an outside link is the case where a new tab is most often wanted.

sridharkalaibala:
Closes #1125.

Sidebar destinations outside the router currently need a custom anchor or a button-based navigation workaround. Add `href` to `SidebarItem` and `SidebarRailItem`, rendering the existing link content as a native `<a>`. This supports normal browser link actions for other apps, documentation, and mail addresses. Frappe's merged Desk shell explicitly copies the rail tile classes around anchors pending this capability ([consumer PR](https://github.com/frappe/frappe/pull/42586)).

```vue
<SidebarRailItem label="Docs" href="https://docs.frappe.io" icon="lucide-book" />
<SidebarItem label="ERPNext" href="/apps/erpnext" />
```

An explicitly supplied `href` takes precedence over `to`, including an empty href. Native sidebar links use explicit `active` rather than inferring state from an ignored route. Existing router links, buttons, slots, tooltips, indicators, and click handlers retain their paths. The component guides and generated API tables document the new prop.

Validation:
- Both Cypress component suites pass in headless Electron: **42 tests**. All ten new cases fail on unchanged `main`, while its 32 existing cases pass. Coverage includes absolute/relative/mail/empty hrefs, a simultaneously supplied `to`, keyboard focus, unprevented native click delivery, and use without a router.
- Both `vue-tsc` app/node projects and the Vite build pass. The build retains its large-chunk warning.
- Full props generation completed; only the two affected API tables are included. `git diff --check` passes.
- Full Vitest run: 1,719 passed, two failures. The migration test relying on chmod permission denial also fails on unchanged `main` under Windows. The editor's cold import exceeds its five-second timeout; all 13 editor primitive tests pass separately with a 20-second timeout. No test or timeout configuration is changed in this PR.

For the full docs/unit run, tracked sources were temporarily normalized to LF to avoid existing Windows shebang/JSDoc parsing issues, then restored. This contribution was prepared with AI assistance; the validation above was run locally.


sridharkalaibala:
Closing as superseded by upstream commit `851d6ed0` (`refactor(navigation)!: use route and href for destinations`). Current `main` now exposes `route` and `href` on both `SidebarItem` and `SidebarRailItem`, renders native anchors for `href`, covers route/href precedence in both Cypress suites, and falls back to a plain anchor for string routes when no router is installed.

That upstream navigation API differs from this older branch's `to` naming and href precedence, so resolving these conflicts would reintroduce a superseded API rather than preserve the feature now shipped on `main`.

PR #1168, opened 2026-09-16
fix!: picker open at mount, List select-all, Tabs reset, TabButtons.focus()

netchampfaris:
Four behavior fixes the spec audit turned up, and a pass over the specs so they describe what the components actually ship.

## What changed

**Pickers open when they are mounted open.** All four seeded their open state to `false` and only watched `open` for later changes. Opening them was not enough on its own: the open-time work lived in a watcher that never fired, so a panel mounted open had no draft, no panel id behind `aria-controls`, no focus move for a custom trigger, and no scroll to the bound time. `PickerShell` and `TimePicker` now run that work from `onMounted`.

**List select-all follows items that change in place.** `ListRows` only refreshed the select-all universe when `items` was swapped for a different array, so a push, a splice, an in-place replacement, a mutated id or a changed `rowKey` left it checking rows that were gone. It now watches the identities the rows render with. Your selection is still yours — nothing is pruned.

**A clicked Tab clears when the route leaves its page.** The reset watched a tab value rather than the URL, so opening a child route under the same tab (`/inbox` → `/inbox/42`) never cleared it. It now clears when a navigation lands on a different path. It deliberately does *not* clear on a query-only or hash-only write that keeps the same tab matched: a panel that stores its own page number in the URL writes exactly that, and clearing there threw the user out of the panel they were using. Aborted, cancelled and duplicate navigations keep the click.

**`TabButtons.focus()`** completes the imperative-API contract, exported as `TabButtonsExposed`.

**Specs.** Every row of the audit is corrected against the source: the Dropdown suffix rule (replacement, not append), MultiSelect's accepted loading behavior, input sizes, picker props, HoverCard milliseconds, the toast wrapper, editor kits, generated typography. Two template-ref policy conflicts — the editor's inferred expose with a writable `isEmpty`, and Popover's `contentEl` — are recorded as open, not written away. No exposed member changed.

## Changelog

- **Pickers (breaking, silent):** `DatePicker`, `DateRangePicker`, `DateTimePicker` and `TimePicker` honor `open` at mount. If you passed a constant `open="true"` and relied on it being ignored, the panel now opens — drop the prop or bind it to your own state. Mounting open emits no `update:open`. `open: true` with `disabled: true` shows the panel, matching the controlled path as it already behaved; the imperative `open()` is still a no-op while disabled.
- **List (fix):** the header's select-all follows rows added, removed, replaced or re-keyed in place.
- **Tabs (fix):** a clicked tab clears when a navigation lands on a different path, which covers child routes under the same tab (`/inbox` → `/inbox/42`). It no longer clears on a query-only or hash-only navigation that keeps the same tab matched, so a panel that stores its state in the URL cannot eject the user. The trade: a background `?filter=unread` written from outside the clicked tab leaves it selected — `afterEach` cannot tell the two apart. A matched-tab change with no navigation behind it still clears the click.
- **TabButtons:** `focus(options?)` on the component ref, typed as `TabButtonsExposed`. It focuses the selected option or the first enabled one, skips disabled options, and forwards `FocusOptions`.
- **MultiSelect:** documentation only — the `loading` prop's described behavior now matches what it has always done. No API change.

## Testing

- 47 new component tests. Each fix's regression cases were run red against the old behavior first: 8 for the pickers, 6 for List, 3 for Tabs.
- Combined Cypress gate over the eight affected specs: 261 passing.
- `yarn test --dir src`: 103 files, 1440 tests passing. `yarn type-check` clean. `yarn docs:gen` then `yarn docs:check` matching.

## Open questions, recorded not settled

- Should `open` respect `disabled` on the pickers?
- `DateRangePicker.clearable` is declared with a default of `true` and never read, while the other two pickers use it.
- In SSR, a `<ListHeader>` before `<ListRows>` renders `aria-checked="false"`, because the universe is only known once `ListRows` sets up. Pre-existing; settling it means deriving the universe in `List`.
- Tabs routed by query alone do not work: `useLink` matches on route records and params and ignores query, so triggers at `/mail?folder=inbox` and `/mail?folder=sent` both report active at either URL and the first stays lit. Found while testing this change; it is a gap in how `routeSelected` matches, not in the reset rule.

Coverage: 76.57% (-0.01% vs `main`)

Docs preview: https://ui.frappe.io/pr-preview/pr-1168/


netchampfaris:
Thanks — the Tabs finding was right, and it's fixed. Round 1 answers:

**Tabs: navigation from inside the clicked tab deselects it — fixed** (`32eadae791`)

Exactly as described: pagination inside the clicked panel ejected the user. The click now ends when a navigation lands on a **different path**, which keeps the `/inbox/42` child-route case and drops this one, plus the older `watch(routeSelected)` for a matched-tab change with no navigation behind it. The `failure` guard is unchanged. The changelog bullet is now **The trade** and states the case being given up: a background `?filter=unread` written from outside the clicked tab leaves it selected, because `afterEach` cannot tell that apart from the panel storing its own page number. `spec/tabs.md` records it as accepted with the reason.

While testing this: **tabs routed by query alone do not work at all**. `useLink` matches on route records and params and ignores query, so triggers at `/mail?folder=inbox` and `/mail?folder=sent` both report active at either URL and the first stays lit. That is a gap in how `routeSelected` matches rather than in the reset rule — recorded in `spec/tabs.md`, left for a separate change.

**Mount-time `requestFocus` — fixed** (`6cb103053d`)

Settled by checking what the default trigger does rather than by taste. `PickerShell` passes `:auto-focus="false"` and `Popover` cancels reka's `openAutoFocus`, so a panel on the default trigger never takes focus — not at mount, not on a later open. reka draws no distinction itself (`FocusScope` fires one `mountAutoFocus` whenever the content mounts), so the custom-trigger path was the inconsistent one. `onOpened` now takes a `moveFocus` flag; the mount path passes `false`. Opening later from a custom trigger still moves focus in. Tests pin both halves on all three date pickers, and the claim is no longer untested.

**`ListRows` scan cost — fixed** (`3033858b3b`)

Real, and in two places: `selectAllState` filtered the universe with `selection.includes(value)`, and `isSelected` did the same per rendered row. Both now use a derived `Set`, rebuilt only when `selection` changes. Added the case nothing covered — a selection array mutated in place rather than replaced. The `itemValues` re-map stays: it is what makes the watcher track row identity, which is the bug this PR fixes.

**`open: true` + `disabled: true` — declined for this PR**

Deliberately left as an open contract question. The disagreement is pre-existing on the controlled path: the `props.open` watch already bypassed `setOpen`'s disabled guard, and the mount seed follows it rather than introducing a second rule for one prop. Settling it means changing that controlled path too, which is a separate behavior break and outside the scope this PR was planned against. It is recorded in `spec/date-picker.md` and in the PR's open-questions list.

**Spec dates — declined**

They were correct when written. This review ran at `2026-09-16T22:33Z`, which is `2026-09-17` in the author's timezone; it is now `2026-09-17` in UTC as well.

netchampfaris:
Round 2 answers. Merged current `main` (`3069f6deab`) first, so the tree includes #1173.

**Merge gate — clear.** All eight checks are green on `3033858b3b`, including all three Cypress shards (4m each). They were still pending when you read them.

**`spec/imperative-api.md` counts — declined, the numbers were already right.**

I re-counted with a comment- and string-aware parse of all 458 `.vue` files under `src/`, rather than a text search:

| | calls | typed at call site | typed via annotated const | untyped |
|---|---|---|---|---|
| this branch | 39 | 32 | 1 (`Select`) | 6 |
| `origin/main` | 38 | 31 | 1 | 6 |

41/34 does not hold on either. `git grep -c 'defineExpose<' origin/main -- src` returns exactly 34, which is where that figure most likely comes from — but it counts `.ts` type declarations and three comment lines that merely name the generic form (`TextInput.vue:183`, `SettingsBody.vue:23`, and one in a `types.ts`). The branch is one ahead of `main` because `43005221b7` adds `defineExpose<TabButtonsExposed>` — the 39th call and the 32nd typed one. So this is a counting-method difference, not branch drift.

Your enumeration underneath is right and unchanged: `Editor`, `Popover`, `PageHeaderTarget`, the three suggestion lists, plus `Select` on the object it publishes. 32 + 1 + 6 = 39.

Fixed the real problem the nit points at, in `5fb8c18f6a`: the paragraph did not say what it measured. It now names the scope (`src/**/*.vue`), the commit the count was taken at, and the fact that a plain text search matches four comments too.

**Also noticed while re-counting, not changed** — `spec/imperative-api.md` §1.6 and §1.5 cite line numbers that have since moved (`CalendarPanel.vue:436` → 388, `PickerShell.vue:251` → 226, `Combobox.vue:435` → 456, `MultiSelect.vue:330` → 348) and §1.6 says "six editor list components" where there are three today. §1 is framed as historical audit evidence, so I left it; flagging in case you read it as current.

netchampfaris:
Round 3 answers, and this is the last review round — merging with `beta-release` once CI is green.

**Changelog heading — fixed** (`170ec298c5`). `### Pickers — \`open\` is honored at mount` now reads `(breaking, silent)`, matching `Tree` above it and the `DatePicker family` entry below. You are right that the heading is what a consumer scans, and it was the one place the label was missing.

**Merge gate — waiting on it.** Holding the merge until the three Cypress shards on the head commit finish, for exactly the reason you give: every test behind the four fixes is a Cypress component test, so a green run on an older head is not evidence for this one.

Thanks for the three rounds — the Tabs finding in round 1 was the one that mattered, and it was right.

PR #1171, opened 2026-09-17
fix(ListView): pass options to TabButtons in ListFooter

### LINKED ISSUE #1170: ListFooter page length picker renders nothing: still passes the removed `buttons` prop to TabButtons
## What happens

`ListFooter` renders an empty box where its page length picker should be. The
right side (row count, Load More) works. The left side is blank. No console
warning.

## Why

`experimental/ListView/ListFooter.vue:7` still passes `buttons` to `TabButtons`:

```vue
<TabButtons
  v-model="pageLengthCount"
  :buttons="pageLengthOptions.map((o) => ({ label: o, value: o }))"
/>
```

That prop was removed in two steps:

1. `128851e1c5` (`v1.0.0-beta.10`) added `options` and marked `buttons` as
   `@deprecated`. The alias still worked, so `ListFooter` was fine.
2. `d090649286` (`feat(tab-buttons)!: rename type to variant, add fluid, drop
   deprecated members`, `v1.0.0-beta.43`) dropped the deprecated members.
   `ListFooter` was never migrated.

So `ListFooter` has been broken since `v1.0.0-beta.43`. `types.ts` last carried
`buttons?:` in `v1.0.0-beta.42`.

The failure is silent because `TabButtons.vue:36` sets `inheritAttrs: false`, so
the unknown `buttons` prop is dropped rather than landing on the root element.
`TabButtons.vue:51` then falls back to `props.options ?? []`, the radio group
gets no items, and the picker renders nothing.

`ListFooter` is exported from `frappe-ui/experimental` and documented in
`ListView.api.md`, so this is a public surface. Every consumer that does not
override the `#left` slot loses its page length control.

## Reproduce

```vue
<ListFooter v-model="pageLength" :options="{ rowCount: 20, totalCount: 100 }" />
```

Expected: `20 / 50 / 100` pills on the left.
Actual: nothing on the left.

## Fix

One word, at `experimental/ListView/ListFooter.vue:7`:

```diff
-          :buttons="pageLengthOptions.map((o) => ({ label: o, value: o }))"
+          :options="pageLengthOptions.map((o) => ({ label: o, value: o }))"
```

Still present on `main` as of today.

## Small nit nearby

`ListFooter.vue:54` reads the option list into a `ref`:

```js
const pageLengthOptions = ref(props.options.pageLengthOptions || [20, 50, 100])
```

That runs once, so changing `options.pageLengthOptions` later has no effect. A
`computed` would fix it. Happy to fold this into the same PR or leave it out.

## Versions

- frappe-ui `1.0.0-beta.63`, broken since `1.0.0-beta.43`
- Found while upgrading Frappe Helpdesk from beta.24 to beta.63: frappe/helpdesk#3769

I can send the PR if that is useful.

8rulerstar:
Closes #1170

Hi, thanks for the detailed write-up in #1170. It made this one easy to follow.

`ListFooter` was still passing `buttons` to `TabButtons`, which was renamed to `options` in beta.10. The prop was dropped silently, so the page length picker rendered empty.

- `ListFooter.vue`: `:buttons` → `:options`, and `String(o)` for the label since `TabButton.label` is a string
- `ListFooter.vue`: `pageLengthOptions` is now a `computed`, so changing `options.pageLengthOptions` after mount takes effect
- `ChartColorsPage.vue`: the same stale `:buttons` prop on the docs page, as mentioned in the triage comment

I added `ListFooter.test.ts` to cover this: it checks that the default 20 / 50 / 100 options render, and that changing `pageLengthOptions` updates them. Both tests fail on the code before this fix and pass with it.

`yarn type-check` and `yarn docs:check` pass locally. `yarn test` has 34 failures in `useColorScheme.spec.ts`, but they fail the same way on `main` without this change, so I think it's something in my local setup.

This is my first contribution to frappe-ui, so please let me know if I missed anything. Thank you for taking a look!

## Changelog

None — this restores the page length picker that stopped rendering in beta.10, with no API or documented behavior changes. Happy to add an entry if you'd prefer one.


netchampfaris:
This is already fixed and merged

8rulerstar:
Ah, I missed that it was already fixed on main — thanks for letting me know! Happy to help with something else another time.

PR #1172, opened 2026-09-17
fix: migrate ListFooter page length picker from removed buttons prop to options

### LINKED ISSUE #1170: ListFooter page length picker renders nothing: still passes the removed `buttons` prop to TabButtons

kingmakeruix:
## What
Migrate `ListFooter`'s page length picker from the removed `buttons` prop to `options`, and fix the same stale prop on the docs chart-colors page.

## Why
`buttons` was dropped from `TabButtons` in `d090649286` but `ListFooter` was never migrated. With `inheritAttrs: false` the unknown prop is silently dropped and `props.options ?? []` resolves to `[]`, so the picker renders nothing. Closes #1170.

## How
- `experimental/ListView/ListFooter.vue`: `:buttons` → `:options`, `label: String(o)` to match the `TabButton.label: string` type, and `ref` → `computed` for `pageLengthOptions` so later option changes take effect
- `docs/components/foundations/ChartColorsPage.vue`: same `:buttons` → `:options` migration (`modeButtons` already matches the `TabButton` shape)

## Checklist
- [x] `npm run type-check` passes
- [x] Prettier clean on changed files
- [x] Unit suite: remaining failures reproduce on clean main too (env-related, unrelated to this change)

Note: prepared with AI assistance under my direction.

Coverage: 76.58% (+3.02% vs `main`)


RitvikSardana:
fix the regression test review @kingmakeruix

kingmakeruix:
Done — added experimental/ListView/ListFooter.ssr.test.ts with two SSR regression tests (default and custom page length options). Verified they fail on main (0 pills rendered) and pass with the fix. Thanks for the review!

PR #1183, opened 2026-09-19
fix(editor)!: rename UploadedFile to UploadedMedia so upload() fits uploadFunction

netchampfaris:
## Why the change

frappe-ui's own `upload()` could not be passed to the editor's `uploadFunction`, because the root and `frappe-ui/editor` each exported a different type named `UploadedFile`.

## Special things to note

- **Breaking for `frappe-ui/editor`:** `UploadedFile` is now `UploadedMedia`, with no deprecated alias (ADR-0008). No app in the local census imports it from `frappe-ui/editor`. Gameplan imports the root type, which keeps its name and shape.
- **The root fix is one keyword.** `interface` becomes `type`. TypeScript gives a type alias an implicit index signature, and gives an interface none. The shape does not change, and typo checks on object literals still work.
- **Verified by agents, who wrote the change:** `yarn type-check` passes, full `yarn test` passes (2208 tests), `yarn docs:check` matches. The new regression test fails when the root type is put back to `interface`.

## Change outline

Two public types shared one name, with shapes that do not fit each other:

```diff
  frappe-ui          UploadedFile   { file_name, file_size, file_url, ...optional File fields }
- frappe-ui/editor   UploadedFile   { file_url, width?, height?, [key: string]: unknown }
+ frappe-ui/editor   UploadedMedia  { file_url, width?, height?, [key: string]: unknown }
```

The root type becomes a type alias, so it satisfies the editor's index signature:

```diff
  // src/utils/useFileUpload.ts
- export interface UploadedFile {
+ export type UploadedFile = {
```

What now compiles:

```ts
import { upload } from 'frappe-ui'
import { useEditor } from 'frappe-ui/editor'

useEditor({ uploadFunction: upload })
// before: TS2322 Index signature for type 'string' is missing in type 'UploadedFile'
```

This shows the types only. Bare `upload` stores a private file attached to nothing, so only the uploader can load it. The migration guide shows the safe form: attach to the owner document with `doctype` and `docname`, or pass `private: false`.

Where the change lands:

```text
src/utils/useFileUpload.ts                          # interface -> type alias
src/molecules/editor/
├── extensions/shared/media-upload-types.ts         # owns UploadedMedia
├── extensions/shared/*.ts, useEditor.ts, index.ts  # follow the rename
└── types.test.ts                                   # regression: upload fits UploadFunction
spec/editor.md, docs/content/docs/molecules/editor.md   # contract and docs use the new name
docs/content/docs/changelog.md, migration.md        # breaking entry, migration row, safe upload example
```

## Changelog

`docs/content/docs/changelog.md`, under Unreleased:

- **The editor's `UploadedFile` is now `UploadedMedia` (breaking).** `frappe-ui` and `frappe-ui/editor` both exported a type named `UploadedFile`, with shapes that did not match. The root `UploadedFile` keeps its name and shape, and is now a type alias. So `useEditor({ uploadFunction: upload })` compiles. The docs say bare `upload` stores a private, unattached file, and show how to attach it. A type alias does not take declaration merging, so extend `UploadedFile` with an intersection instead. Code that imports `UploadedFile` from `frappe-ui/editor` renames the import to `UploadedMedia`. The migration guide has the row.

Docs preview: https://ui.frappe.io/pr-preview/pr-1183/

Coverage: 76.83% (±0.00% vs `main`)


netchampfaris:
Fixed all three in 3881bfa459:

- **Bare `upload` stores private, unattached files:** fixed. The migration guide now says so plainly. The example attaches the file to the owner document (`doctype`, `docname`) and passes `signal` and `onProgress` through. `private: false` is the other option, for public content. The changelog bullet links to that section. I kept private as the default because it is a deliberate security choice.
- **Conditional wording at `migration.md:2968`:** fixed. The item now says "Rename the editor's `UploadedFile` import to `UploadedMedia`."
- **`interface` to `type` ends module augmentation:** fixed. A new changelog bullet says so and shows `type MyFile = UploadedFile & { ... }`.

The PR body is updated to match.

netchampfaris:
Fixed in b502d63483: the first `UploadFunction` example at `migration.md:2960` now passes `doctype` and `docname`, like the example below it. Both snippets type-check against `UploadFunction`.

netchampfaris:
@greptileai review

